PR #1, opened 2025-04-15
Prototype of architecture. Auth is broken and disabled.

tdresser:
I'd love your feedback on the direction here!


.gitignore
```diff
@@ -1,136 +1,3 @@
-# Logs
```
tdresser, line 1: Happy to keep these, or just add them back as needed.

docker/compose.yml
```diff
@@ -0,0 +1,62 @@
...
+      - app_network
+    restart: unless-stopped
+    healthcheck:
+      test: ['CMD', 'pg_isready', '-U', 'supertokens_user', '-d', 'grassroots']
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+  supertokens:
+    container_name: supertokens
```
IanEdington, line 25: Super token looks really cool. I'd like to learn more about what we can or can't do at the free tier. I have dreams of setting up a unified sign-in experience for volunteers and wouldn't want to lock us into a platform that would make that prohibitively expensive.

.husky/pre-commit
IanEdington, line 1: I'll try this out when I spin this up locally, but I'm interested in how long it takes to run all of this in the pre-commit hook. I've seen how frustrated people become with pre-commit hooks and I'd like to avoid having them be an annoyance.

eslint.config.mjs
```diff
@@ -0,0 +1,69 @@
...
+  },
+  {
+    rules: {
+      '@typescript-eslint/no-floating-promises': 'error',
+      '@typescript-eslint/no-unsafe-argument': 'error',
+      '@typescript-eslint/require-await': 'error',
+      '@typescript-eslint/no-non-null-assertion': 'off',
+      // This is a bit annoying, but worth it to ensure controller routes are typed correctly.
+      // Or maybe we should only turn it on for controllers?
+      // tdresser@: thinks it's probably fine to leave it on globally.
```
IanEdington, line 61: Let's leave it on until it becomes annoying :)

.husky/pre-commit
tdresser: These definitely could be part of git ci, not precommit.

eslint is surprisingly slow - it's the main thing I'm not sure about.

prettier is only running on changed files due to lint-staged, and the others are pretty quick, even on the low power intel NUC I'm developing on.

Definitely give it a shot though! I don't have any experience using github's CI tools, do you think that's a reasonable thing to migrate to if this is too slow?

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "all"
```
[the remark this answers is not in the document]
tdresser: I _think_ I pulled this from the default prettier config nest.js created when I used it's CLI.

I also don't care at all which options we use. I'll switch to an empty config and apply that style.

docker/caddy/Caddyfile
```diff
@@ -0,0 +1,14 @@
+http://grassroots.local {
```
tdresser, line 1: In case you're curious, this isn't being used at all right now, but I expect we'll need it for getting auth working consistently between local dev and a hypothetical production environment.

docker/compose.yml
tdresser: I _think_ the free tier mostly just means you need to host it yourself. It doesn't seem super constrained.

Though there was something about even when you're hosting it yourself, their "use management dashboard" is heavily restricted, to 3 users or something.

Definitely worth double checking.

docker/caddy/Caddyfile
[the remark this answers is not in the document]
tdresser: I'll take a look!

docker/compose.yml
tdresser: Oh, there are a bunch of things that cost money even when you self-host.
https://supertokens.com/pricing has the details.

I think you've got much more context than I do on how crucial those features are, and how reasonable the pricing is.

.husky/pre-commit
IanEdington: yeah absolutely. We're running our other CI on GitHub Actions, and it's working well.

PR #2, opened 2025-04-17
Bootstrap nest

tdresser:
This CL just contains the result of running the nest CLI to generate a new app.

```sh
nest new grassroots-backend 
```

I selected npm as package manager, despite believing that pnpm is strictly better, because npm is somewhat more widely used.

I'd be extremely happy to use pnpm instead, if folks are fine with that option.

### DIFF
--- a/LOG.md
+++ b/LOG.md
@@ -1,0 +1,7 @@
+# Bootstrapping Nest
+Shortly we'll move the environment into docker, but to bootstrap, we'll just install the nest CLI globally.
+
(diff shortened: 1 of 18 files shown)


IanEdington:
I agree we should stick with npm unless we run into an issue that pushes us towards pnpm... but I also like pnpm :D

IanEdington (approved):

IanEdington:
LGTM!

PR #5, opened 2025-04-23
Build an App Shell with Tanstack Router and Mantine

tdresser:
- Remove unused Vite Assets
- Set up Mantine according to [these instructions](https://mantine.dev/guides/vite/).
- Set up Tanstack Router according to [these instructions](https://tanstack.com/router/latest/docs/framework/react/quick-start).

For an example of where Tanstack Router provides stronger typing guarantees than other routers, see [here](https://github.com/gpo/grassroots/blob/7404242ec967dff3cd7373efb4baddd4e7093da2/README.md#routes). This is the primary reason to go with Tanstack Router over alternatives (though it has some other ergonomics benefits).


grassroots-frontend/src/routeTree.gen.ts
```diff
@@ -0,0 +1,134 @@
+/* eslint-disable */
+
+// @ts-nocheck
+
+// noinspection JSUnusedGlobalSymbols
+
+// This file was automatically generated by TanStack Router.
+// You should NOT make any changes in this file as it will be overwritten.
+// Additionally, you should also exclude this file from your linter and/or formatter to prevent it from being checked or modified.
```
IanEdington, line 9: If you add a `.gitattributes` file with the following, github will collapse this file by default.

```
**/*.gen.ts linguist-generated=true
```

grassroots-frontend/src/routes/index.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export const Route = createFileRoute('/')({
+  component: Index,
+})
+
+function Index() {
+  return (
+    <p>TODO</p>
+  )
+}
```
IanEdington, line 11: Based on the file endings, I think this might not have been prettified

IanEdington (commented):
Looks great! I'm loving what I'm reading about tanstack router. It seems like a huge leap over react-router in terms of useful functionality and typesafety.

Formatting with prettier is the only thing I think should block this PR

IanEdington (commented):
Looks great! I'm loving what I'm reading about tanstack router. It seems like a huge leap over react-router in terms of useful functionality and typesafety.

Formatting with prettier is the only thing I think should block this PR

grassroots-frontend/src/routeTree.gen.ts
```diff
@@ -0,0 +1,134 @@
...
+      "filePath": "index.tsx"
+    },
+    "/create-contact": {
+      "filePath": "create-contact.tsx"
+    },
+    "/search": {
+      "filePath": "search.tsx"
+    }
+  }
+}
```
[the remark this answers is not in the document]
tdresser: Yeah... kinda weird. I wonder if the vite plugin uses it?

grassroots-frontend/src/routeTree.gen.ts
tdresser: Done!

grassroots-frontend/src/routes/index.tsx
tdresser: Done!

IanEdington (approved):

PR #8, opened 2025-04-27
Use TypeORM and Postgres in Contacts Service

tdresser:
There were a bunch of gotchas here, from the tsconfig allowing absolute imports, but jest not supporting them, to cryptic errors from jest running tests in parallel.

I've tried to call out these out clearly.

### DIFF
--- a/LOG.md
+++ b/LOG.md
@@ -3,6 +3,11 @@
 Going with postgres (pg).
 Docker compose from [here](https://hub.docker.com/_/postgres).
 
+A few notes:
+
+- Use of rolled back transactions in tests seems to be best practice, but IDs appear to increment in rolled back transactions.
+-
+
 # Installing Tanstack Router
 
 Followed [these instructions](https://tanstack.com/router/latest/docs/framework/react/quick-start).
--- a/LOG.md
+++ b/LOG.md
@@ -15,7 +15,6 @@ Followed [these instructions](https://tanstack.com/router/latest/docs/framework/
 # Installing Mantine
 
 Followed [these instructions](https://mantine.dev/guides/vite/).
-npm install ts-jest-resolver --save
 
 # Bootstrapping Vite
 
--- a/LOG.md
+++ b/LOG.md
@@ -6,7 +6,7 @@ Docker compose from [here](https://hub.docker.com/_/postgres).
 A few notes:
 
 - Use of rolled back transactions in tests seems to be best practice, but IDs appear to increment in rolled back transactions.
--
+- We've switched to running all tests sequentially, which will be bad for test performance on devices with a reasonable core count. Eventually we should try to reparallelize these.
 
 # Installing Tanstack Router
 
--- a/LOG.md
+++ b/LOG.md
@@ -1,3 +1,8 @@
+# Pulling in TypeORM
+
+Going with postgres (pg).
+Docker compose from [here](https://hub.docker.com/_/postgres).
+
 # Installing Tanstack Router
 
 Followed [these instructions](https://tanstack.com/router/latest/docs/framework/react/quick-start).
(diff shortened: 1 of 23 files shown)


grassroots-backend/src/contacts/contacts.service.spec.ts
```diff
@@ -1,18 +1,56 @@
...
+  it("should create and return a contact", async () => {
+    const contact: CreateContactInDto = {
+      email: "test@test.com",
+      firstName: "Test",
+      lastName: "Test",
+      phoneNumber: "999-999-9999",
+    };
+    const created = await service.create(contact, queryRunner);
+
+    const allContacts = await service.findAll(queryRunner);
```
IanEdington, line 50: 🙌

grassroots-backend/src/contacts/contacts.controller.spec.ts
```diff
@@ -1,17 +1,30 @@
...
+  afterAll(async () => {
+    await app.close();
+  });
 
   beforeEach(async () => {
-    const module: TestingModule = await Test.createTestingModule({
-      controllers: [ContactsController],
-      providers: [ContactsService],
-    }).compile();
+    await startTestTransaction();
```
IanEdington, line 23: 😎

grassroots-backend/src/testing/dbTransactions.ts
```diff
@@ -0,0 +1,20 @@
+import { DataSource, QueryRunner } from "typeorm";
+import { getTestApp } from "./getTestApp";
+
+let queryRunner: ReturnType<DataSource["createQueryRunner"]>;
+
+export async function startTestTransaction(): Promise<QueryRunner> {
+  const dataSource = (await getTestApp()).get(DataSource);
+  queryRunner = dataSource.createQueryRunner();
+  await queryRunner.connect();
+  await queryRunner.startTransaction();
```
IanEdington, line 10: I don't think it should block this PR. I'm wondering if we use transactions in our application, will this have an adverse effect?

dev.env
IanEdington, line 1: I'm not used to committing `.env` files since they often contain local secrets / variables.
I've oven seen an ignore on the `.env` file and a `.env.sample` with the default starting values.
A step in the readme is often to copy the `.env.sample` to `.env` and fill in the required values.

LOG.md
```diff
@@ -1,3 +1,13 @@
+# Pulling in TypeORM
+
+Going with postgres (pg).
+Docker compose from [here](https://hub.docker.com/_/postgres).
+
+A few notes:
+
+- Use of rolled back transactions in tests seems to be best practice, but IDs appear to increment in rolled back transactions.
+- We've switched to running all tests sequentially, which will be bad for test performance on devices with a reasonable core count. Eventually we should try to reparallelize these.
```
IanEdington, line 9: non-blocking: Is it theoretically possible to parallelize tests using the transaction rollback strategy?

IanEdington (approved):
Nothing blocking LGTM!

You probably want to merge before I do since it'll be easier to rebase my changes on this PR

LOG.md
tdresser: Maybe! You definitely need to avoid modifying the schemas in parallel, but in theory, I think this should probably be doable.

Not sure!

Filed https://github.com/gpo/grassroots/issues/16

grassroots-backend/src/testing/dbTransactions.ts
tdresser: I think it would be manageable, but could get a bit clunky.
Worst case, if we're running tests, we run everything outside a transaction, which makes the tests less representative.

Postgres does have some nested transaction support, but I'm not sure how typeorm interacts with it.

Another option would be to have a "leave transactions open" param on each of our service methods that start transactions. Then the test can manually rollback the transactions when appropriate.

Happy to dig in here - thoughts on the priority?

PR #11, opened 2025-04-28
feat: consolidate linting at project root

IanEdington:
consolidate eslint and prettier at the root level project


eslint.config.mjs
```diff
@@ -0,0 +1,58 @@
...
+      ecmaVersion: 5,
+      sourceType: "module",
+      parserOptions: {
+        projectService: true,
+        tsconfigRootDir: import.meta.dirname,
+      },
+    },
+    settings: {
+      react: {
+        version: "detect",
```
tdresser, line 39: Optional: this complains about not having a version of react present in this package.json.

It might be worth adding React here just to ensure this is using the right version of the react linting rules? Then syncpack can ensure that the react versions are consistent.

tdresser (approved):
This is great, thanks!

eslint.config.mjs
IanEdington: Thanks, I've added it to the root-level package.json

PR #12, opened 2025-04-28
fix: existing eslint error

IanEdington:
related to #11

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -53,6 +53,7 @@ export default tseslint.config(
         "error",
         { assertionStyle: "never" },
       ],
+      "@typescript-eslint/no-misused-spread": "off",
     },
   },
 );
--- a/grassroots-frontend/src/main.tsx
+++ b/grassroots-frontend/src/main.tsx
@@ -15,6 +15,7 @@ declare module "@tanstack/react-router" {
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 createRoot(document.getElementById("root")!).render(
   <StrictMode>
     <RouterProvider router={router} />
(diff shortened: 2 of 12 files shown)


grassroots-frontend/src/main.tsx
```diff
@@ -15,6 +15,7 @@ declare module "@tanstack/react-router" {
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 createRoot(document.getElementById("root")!).render(
```
IanEdington, line 19: I've found this to be a useful error to have. It adds a little friction when using the `!`. Enough to stop its unnecessary use, but not enough to stop legitimate uses like the above.

tdresser: The other approach I've used (before I knew ! existed) is to define:

```javascript
export function el(id: string, message: string | undefined): Element {
  const result = document.getElementById(id);
  if (result) {
    return result;
  }
  throw new Error(message ?? `Couldn't find element with id ${id}`);
}
```

I don't find myself using ! in many other places. I'm happy either way.

tdresser (approved):
LGTM!

eslint.config.mjs
```diff
@@ -53,6 +53,7 @@ export default tseslint.config(
         "error",
         { assertionStyle: "never" },
       ],
+      "@typescript-eslint/no-misused-spread": "off",
```
IanEdington, line 56: @tdresser I wanted to flag a change I made to the ESLint config. This rule blocks dangerous spreads, including class objects. While it would be useful to block things like functions and strings, I couldn't find a way to allow only class objects to be spread.

https://github.com/gpo/grassroots/blob/5799501c5506c9ca4e0087e061ac5f4a3cbf60af/grassroots-backend/src/contacts/contacts.service.spec.ts#L52-L54

tdresser: You can always use instanceToPlain from class transformer, and then spread that.

IanEdington: ok, sounds good. Let's put this back in

PR #19, opened 2025-04-29
Create 100 random contacts

tdresser:
This will help with testing search.

It also verifies that the rollback based test approach works fine - transactions are nested properly.


grassroots-backend/src/contacts/contacts.service.ts
```diff
@@ -35,4 +36,27 @@ export class ContactsService {
     const repo = getRepo(ContactEntityOutDTO, queryRunner, this.dataSource);
     return await repo.findOneBy({ id });
   }
+
+  async addFakesToDatabase(
```
IanEdington, line 40: I worry about having test code so tightly integrated into production code. I think it would be less error prone and easier for new devs to do the right thing if it was in a separate `test` directory.

IanEdington:
I changed the branch to OpenAPI so it shows only the commits from here

image

grassroots-backend/src/contacts/contacts.service.ts
tdresser: Oh, I think I misinterpreted. I actually want a button in the UI for testing the search functionality in the frontend. 

Not just for automated tests.

Do you think code like that should live in a separate test directory?

This would obviously not exist in production.

Or maybe this should be a script, not a button?

tdresser: Idea: maybe I should just have a separate controller with all Dev related things? And then it only gets injected in "dev mode"?

IanEdington: interesting. Is it kind of like devtools for the application? It would be a valid API in dev environments but not in production.

Concerns:
* accidentally creating many bogus contacts in production
* accidentally using code intended for tests in production

Thoughts:
* The way I would approach getting data into development environments (not necessarily the best way) would be to generate seed data in an initialization step.
* Could we have an environment variable with the current execution environment, something like `prod`, `stage`, `local`. We could add an annotation guard on endpoints `@DevelopmentOnly` or something similar? That way, the endpoints with this guard wouldn't be able to execute?
* I've seen it before where there's an `is_test` flag on the generated objects. I find this strategy error prone since you have to remember to exclude those records every time you query them.

I like the idea of a dev mode controller.
If the code for the dev mode is all together, it should be relatively easy to exclude it via eslint rule. (In the future when it becomes a problem)

PR #21, opened 2025-05-01
Provide form for creating contacts

tdresser:
This is currently only verified manually, via logging in the backend, which isn't great. Once Search exists, manual testing will be trivial, and I think that's likely sufficient at this stage?

Note that I switched the format of env files to what Vite wants: .env.development and .env.production.
To expose environment variables to the frontend, they must be prefixed with VITE\_, and added to vite-env.d.ts. This prevents accidentally leaking things like passwords to the frontend.

Validation will come in a followup PR.


grassroots-frontend/src/grassRootsAPI.ts
```diff
@@ -0,0 +1,6 @@
+import createFetchClient from "openapi-fetch";
+import { paths } from "./grassroots-shared/openAPI.gen";
+
+export const grassrootsAPI = createFetchClient<paths>({
+  baseUrl: import.meta.env.VITE_BACKEND_HOST,
+});
```
IanEdington, line 6: 🤩🤩🤩

grassroots-frontend/src/main.tsx
```diff
@@ -4,20 +4,28 @@ import { RouterProvider, createRouter } from "@tanstack/react-router";
...
 // Import the generated route tree
 import { routeTree } from "./routeTree.gen";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 
 // Create a new router instance
 const router = createRouter({ routeTree });
 
+console.log(import.meta.env.VITE_BACKEND_HOST);
+console.log(import.meta);
+
```
IanEdington, line 14: Are these meant to be here?

grassroots-frontend/src/routes/create-contact.tsx
```diff
@@ -1,10 +1,66 @@
...
+    },
+    [],
+  );
+
+  return (
+    <FormProvider {...form}>
+      {/* This little typescript dance is required to make eslint happy.  */}
+      <form onSubmit={(...args) => void form.handleSubmit(onSubmit)(...args)}>
+        <TextInput
+          id="firstName"
```
IanEdington, line 52: Why is ID needed on this element but not others?

IanEdington (approved):
Wow! This looks amazing!

grassroots-frontend/src/main.tsx
tdresser: Nope, removing.

grassroots-frontend/src/routes/create-contact.tsx
tdresser: It's not, removing.

PR #22, opened 2025-05-01
Implement and test validated/transformed casts

tdresser:
To use class-validator and class-transformer in frontend code, we need a way to take a set of inputs (.e.g., form fields) and apply class-validator / class-transformer to create an instance of the appropriate class.

This PR supplies two such methods, one which assumes the inputs are of the appropriate type, and the other of which magically converts the inputs the the right type, (e.g., magically applying `parseInt` if needed.

### DIFF
--- a/grassroots-backend/src/grassroots-shared/cast.ts
+++ b/grassroots-backend/src/grassroots-shared/cast.ts
@@ -1,0 +1,43 @@
+import { ClassConstructor, plainToInstance } from "class-transformer";
+import { validateSync } from "class-validator";
+
+// Needs to be imported somewhere for `enableImplicitConversion` to work.
+import "reflect-metadata";
+
+// Roughly, given a class, gives you an interface with its attributes.
+export type PropsOf<T> = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+  [K in keyof T as T[K] extends Function ? never : K]: T[K];
+};
+
+// Roughly, given a class, gives you an interface with the same attributes, but all untyped.
+export type PropsOfAsUnknown<T> = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+  [K in keyof T as T[K] extends Function ? never : K]: unknown;
+};
+
+export function cast<T extends object>(
+  cls: ClassConstructor<T>,
+  plain: PropsOf<T>,
+): T {
+  const instance = plainToInstance(cls, plain);
+  const validationErrors = validateSync(instance);
+  if (validationErrors.length > 0) {
+    throw new Error(validationErrors.join("\n"));
+  }
+  return instance;
+}
+
+export function castWithConversion<T extends object>(
+  cls: ClassConstructor<T>,
+  plain: PropsOfAsUnknown<T>,
+): T {
+  const instance = plainToInstance(cls, plain, {
+    enableImplicitConversion: true,
+  });
+  const validationErrors = validateSync(instance);
+  if (validationErrors.length > 0) {
+    throw new Error(validationErrors.join("\n"));
+  }
+  return instance;
+}
--- a/grassroots-backend/src/testing/typetests/cast.test.ts
+++ b/grassroots-backend/src/testing/typetests/cast.test.ts
@@ -1,0 +1,39 @@
+import * as tstyche from "tstyche";
+import { cast, castWithConversion } from "../../grassroots-shared/cast";
+import { IsNumber, Min } from "class-validator";
+
+class Test {
+  @IsNumber()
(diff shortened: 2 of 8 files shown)


grassroots-backend/src/testing/typetests/cast.test.ts
```diff
@@ -0,0 +1,42 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
tdresser, line 2: So here's a fun puzzle...

Without this, the commit checks don't pass, due to the type errors in this file.

However, with this, the tests no longer pass!

Excluding the file from the tsconfig also causes the test to fail.

Any thoughts?

tdresser: Aha, got it.

So I've got separate tests for type assertions and functional testing, and a tsconfig.json specifically for the type assertion test.

The type assertion test is then excluded from the tsconfig.json that we actually typecheck. A bit annoying, but works fine!

grassroots-backend/src/grassroots-shared/cast.ts
```diff
@@ -0,0 +1,43 @@
...
+import "reflect-metadata";
+
+// Roughly, given a class, gives you an interface with its attributes.
+export type PropsOf<T> = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+  [K in keyof T as T[K] extends Function ? never : K]: T[K];
+};
+
+// Roughly, given a class, gives you an interface with the same attributes, but all untyped.
+export type PropsOfAsAny<T> = {
```
IanEdington, line 14: would this be more accurately `PropsOfAsUnknown`?

grassroots-backend/src/testing/typetests/cast.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  @Min(0)
+  a!: number;
+
+  constructor(a: number) {
+    this.a = a;
+  }
+}
+
+describe("cast", () => {
+  it("should cast in the trival case", () => {
```
IanEdington, line 16:
```suggestion
  it("should cast in the trivial case", () => {
```

IanEdington (approved):
Looks awesome!

grassroots-backend/src/grassroots-shared/cast.ts
tdresser: Whoops, thanks. Fixed.

grassroots-backend/src/testing/typetests/cast.test.ts
tdresser: Thanks, fixed.

grassroots-backend/src/testing/typetests/cast.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  it("should throw a type error if the properties don't overlap", () => {
+    tstyche
+      .expect(cast(Test, { b: 3 }))
+      .type.toRaiseError("does not exist in type");
+  });
+
+  it("should throw a type error if the types don't overlap", () => {
+    tstyche
+      .expect(cast(Test, { a: "2" }))
```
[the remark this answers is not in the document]
tdresser: Yeah, this is super whacky. Details are here: https://tstyche.org/guide/introduction#how-tstyche-works

This test is never run. Their examples include files which are both run and checked by tstyche, but I didn't figure that was worth figuring out, so we just have two files, one for type checks (which is never run), and one with functional tests.

grassroots-backend/src/testing/typetests/cast.test.ts
```diff
@@ -0,0 +1,39 @@
...
+describe("cast", () => {
+  it("should cast in the trivial case", () => {
+    const testCasted = cast(Test, { a: 2 });
+    tstyche.expect(testCasted).type.toBe<Test>();
+  });
+
+  it("should throw a type error if the properties don't overlap", () => {
+    tstyche
+      .expect(cast(Test, { b: 3 }))
+      .type.toRaiseError("does not exist in type");
```
mrazauskas, line 24: @tdresser Glad to see you found TSTyche useful. I am the author of the library.

To draw your attention, now you can use the `.toBeCallableWith()` matcher instead of expecting errors. This and other ability matchers ship with `tstyche@next`. This is a beta release, but there will be nothing breaking for CLI usage anymore.

The assertion would read like this:

```ts
expect(cast).type.not.toBeCallableWith(Test, { b: 3 });
```

There is no error, so you can move it back to the functional test file.

---

Removing `.not` gives this error:

Screenshot 2025-05-05 at 07 45 21

PR #24, opened 2025-05-02
Remove metadata.ts from repo

tdresser:


IanEdington (approved):

IanEdington:
I've noticed metadata is a persistent problem with PRs. Is there a reason this hasn't been merged?

If there's some reason we can't remove `metadata.ts` from source, we could prettify the file so its diffs are cleaner and hopefully less likely to cause merge conflicts. This comes with the tradeoff of always having a diff on `metadata.ts` during development, which seems to be the case now anyways.

PR #25, opened 2025-05-02
Validate the create form.

tdresser:
Validating the create form requires getting mantine and react-hook-form to play nicely together. We likely could have used https://github.com/aranlucas/react-hook-form-mantine, but it hasn't been updated in years, and rolling our own strategy here doesn't seem too painful.

Note: This doesn't actually depend on cast, as I had thought it did. For this, we can just use @hookform/resolvers/class-validator.


grassroots-frontend/src/components/textField.tsx
```diff
@@ -0,0 +1,44 @@
...
+// when used in this way. Explicitly make it required.
+export type TextFieldProps<T extends FieldValues> = Omit<
+  UseControllerProps<T>,
+  "defaultValue"
+> &
+  Required<Pick<UseControllerProps<T>, "defaultValue">> &
+  Omit<TextInputProps, "value" | "defaultValue">;
+
+export function TextField<T extends FieldValues>(
+  props: TextFieldProps<T>,
```
IanEdington, line 21: This type seems like it could be 

```
{
name: ...
defaultValue: ...
label: ...
}
```

Should `props` be passed into the `TextInput`?

grassroots-frontend/src/components/textField.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <TextInput
+        value={value}
+        onChange={(v) => {
+          onChange(v.target.value);
+        }}
+        label={props.label}
+        error={fieldState.error?.message}
+        {...field}
+      ></TextInput>
+    </>
```
IanEdington, line 42: Is the fragment needed?

IanEdington (approved):

grassroots-frontend/src/components/textField.tsx
tdresser: Yeah, the approach I took here is both broken and incorrect!

We probably want to narrow the API surface of this as much as possible, so taking TextFieldProps is incorrect.

grassroots-frontend/src/components/textField.tsx
tdresser: Nope. Added https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md to avoid this in the future.

grassroots-frontend/src/components/textField.tsx
tdresser: Fixed.

eslint.config.mjs
```diff
@@ -45,15 +45,13 @@ export default tseslint.config(
...
-      // This is a bit annoying, but worth it to ensure controller routes are typed correctly.
-      // Or maybe we should only turn it on for controllers?
-      // tdresser@: thinks it's probably fine to leave it on globally.
       "@typescript-eslint/explicit-function-return-type": "error",
       "@typescript-eslint/consistent-type-assertions": [
         "error",
         { assertionStyle: "never" },
       ],
       "@typescript-eslint/no-misused-spread": "off",
+      "react/jsx-no-useless-fragment": "error",
```
IanEdington, line 54: 👌

PR #26, opened 2025-05-03
feat(CI): add eslint to CI

IanEdington:
Run eslint as a PR check

### Fixable Errors
eslint provides a mechanism for fixing some errors. 2bf9394d is an example of a modification commit if the eslint error is fixable.

### Annotations
When eslint errors are not fixable they will be added to the PR as annotations, like in [`0bc0ef6` (#26)](https://github.com/gpo/grassroots/pull/26/commits/0bc0ef67d9f35cd980da6609b2d25b8fd565df1f)

image


.gitignore
```diff
@@ -8,5 +8,8 @@ grassroots-frontend/vite.config.js
 .env.test
 .env
 
+# CI
+eslint_report.json
+
 # IDE
-.idea
+.idea
```
IanEdington, line 15:
```suggestion
.idea

```

PR #28, opened 2025-05-03
feat(CI): add syncpack to CI checks

IanEdington:
This PR runs syncpack as part of our blocking CI checks. It will ensure packages are in sync across the root, backend, and frontend directories.

### Fixable
Since syncpack is fixable, this is added to the commit modification process.https://github.com/gpo/grassroots/pull/28/commits/91e56059bd678a69ae7c2a4c57b441bc2da94e70 is an example of a modification commit.

### Move syncpack from `pre-commit` to `lintstaged`
Only run syncpack when a package.json or package-lock.json file is modified
image


.lintstagedrc.cjs
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  "!(*.ts|*.tsx)": "prettier --write --ignore-unknown",
+  "*.{ts,tsx}": ["eslint --fix", "prettier --write"],
+  "*{package.json,package-lock.json}": () => ["syncpack list-mismatches"],
```
IanEdington, line 4: moving to a `js` file allows running files that don't take filenames

.github/workflows/ci.yml
```diff
@@ -37,6 +37,13 @@ jobs:
           npm install --prefix grassroots-frontend
           npm install --prefix grassroots-backend
 
+      - name: Run syncpack
+        run: |
+          npx syncpack fix-mismatches
```
IanEdington, line 42: Use `fix-mismatches` rather than `list-mismatches` since we can commit modified files

.lintstagedrc.cjs
tdresser: Ah, nice!

.github/workflows/ci.yml
tdresser: Just confirming - if fix-mismatches runs and creates a new commit, the whole pipeline runs again, right?

We want to avoid the case where fix-mismatches performs a package version bump that breaks things, after we've run the tests.

tdresser (approved):
LGTM, % confirmation that tests will run after a fix commit occurs.

.github/workflows/ci.yml
IanEdington: I'm not confident I understand what you mean, but I'll give an example I think will answer your question.

# commit 1
Let's say we push commit1, which updates a package in backend/package.json and puts it out of sync with frontend/package.json.
Tests pass on commit 1
Syncpack fails and will update frontend/package.json + lock file to the new version and push commit 2.
Overall commit 1 fails CI, because of packages being out of sync

# commit 2 (from CI)
both syncpack and tests will re-run on commit 2
If the package update causes tests to fail, commit 2 will fail the CI check
Overall commit 2 fails CI, because of failed tests, due to updated packages.

# conclusion
Everything gets re-checked for the second commit

tdresser: Perfect!

PR #30, opened 2025-05-04
chore: check symlink into source code

IanEdington:
As discussed in Slack, we'll check in the symlink in order to avoid having to debug issues where the symlink doesn't exist.

### DIFF
--- a/grassroots-frontend/src/grassroots-shared
+++ b/grassroots-frontend/src/grassroots-shared
@@ -1,1 +1,1 @@
-/Users/ian/code/gpo/grassroots/grassroots-backend/src/grassroots-shared/
+../../grassroots-backend/src/grassroots-shared/


grassroots-frontend/src/grassroots-shared
```diff
@@ -0,0 +1 @@
+/Users/ian/code/gpo/grassroots/grassroots-backend/src/grassroots-shared/
```
tdresser, line 1: This needs to be a relative path, but symlinks use absolute paths, right? Tricky.

tdresser: Once we dockerize, this can be an absolute path.

IanEdington: Oops. Thanks for catching that! We should be able to make it a relative path.

tdresser (approved):

PR #31, opened 2025-05-04
Enable backend validation

tdresser:
This was much gnarlier than I expected. A few things going on here:
- The NestJS OpenAPI CLI plugin doesn't include the schema for the validation error response, which kinda makes it useless. Should I just pull the plugin completely? We might want some kind of linting for ensuring that an APIResponse decorator exists for each controller route.
- Test code previously wasn't initializing the App propertly, so I've refactored that to allow shared App initialization logic between tests and production.
- The "App" controller was returning a raw string, which openapi-fetch doesn't like. I've switched to have that return an object, so we can consistently use openapi-fetch.
- And finally, I added `app.useGlobalPipes(new ValidationPipe({ transform: true }));`, which I thought was all this patch would require!


grassroots-backend/src/decorators/apiValidationErrorResponse.ts
```diff
@@ -0,0 +1,29 @@
...
+      description: "Validation failed",
+      schema: {
+        type: "object",
+        properties: {
+          statusCode: { type: "number" },
+          message: {
+            type: "array",
+            items: { type: "string" },
+          },
+          error: { type: "string" },
```
IanEdington, line 24: Is this meant to be a generic error response schema? It might be difficult to create a consistent error schema all errors will follow.

IanEdington (approved):

IanEdington (approved):
I like this a lot better than the annotation implementation.

PR #32, opened 2025-05-06
Ensure all OpenAPI Values are typed

tdresser:
It's easy to accidentally not correctly type an endpoint.

This currently just throws an exception at runtime – ideally this might be done outside of runtime, but this was the lowest effort, and I really want to know as soon as I write an endpoint incorrectly that it isn't typed right.

I also fixed the one untyped path.


grassroots-backend/src/contacts/contacts.controller.ts
```diff
@@ -21,8 +22,14 @@ export class ContactsController {
...
 
+  @ApiOkResponse({
+    description: "Returns a contact or null if not found",
+    schema: {
+      oneOf: [{ $ref: getSchemaPath(ContactEntityOutDTO) }, { type: "null" }],
+    },
+  })
   @Get(":id")
-  findOne(@Param("id") id: string): Promise<ContactEntityOutDTO | null> {
+  findOne(@Param("id") id: number): Promise<ContactEntityOutDTO | null> {
```
IanEdington, line 32: Interesting that this doesn't work as expected

IanEdington (approved):
Be careful not to merge this until `enable_backend_validation` has been merged. Otherwise, you'll loose the distinct commits on the main branch.

grassroots-backend/src/contacts/contacts.controller.ts
tdresser: I intentionally switched from string to number just to match the database schema. It was working fine, though there was an ugly cast which I've now gotten rid of.

PR #44, opened 2025-05-14
chore: remove prettier from eslint

IanEdington:
prettier is running on it's own and doesn't also need to be run inside eslint


tdresser (commented):
Do we want to use https://github.com/prettier/eslint-config-prettier to avoid conflicts between prettier and eslint?

tdresser:
Looks like I added it as a dependency, but never used it...

Either way, shouldn't block.

tdresser (approved):
LGTM with optional thought above.

PR #46, opened 2025-05-15
Remove no-index check, as tanstack router requires a file named index.

tdresser:


IanEdington:
could we make an exception for tanstack routers index file? I still think this is generally a good rule, it just has an exception

IanEdington (commented):
Could we make an exception for tanstack routers index file? I still think this is generally a good rule, it just has an exception

tdresser:
Will do (when I get the chance)

PR #49, opened 2025-05-15
Implement bulk create

tdresser:
We'll need this eventually, and I want to use it for https://github.com/gpo/grassroots/pull/19, making it easy to test search etc in dev mode.

### DIFF
--- a/grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
+++ b/grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
@@ -21,8 +21,12 @@ export class CreateContactInDto {
   phoneNumber!: string;
 }
 
-export class CreateContactInDtoArray {
+export class CreateBulkContactRequestDto {
   contacts!: CreateContactInDto[];
+}
+
+export class CreateBulkContactResponseDTO {
+  ids!: number[];
 }
 
 @Entity()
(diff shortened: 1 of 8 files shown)


grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
```diff
@@ -21,6 +21,10 @@ export class CreateContactInDto {
   phoneNumber!: string;
 }
 
+export class CreateContactInDtoArray {
```
IanEdington, line 24: My only question is on the use of `Array` in our naming.

Rather than a database `In` class, this seems more like a request object that has an array of contacts on it, but also might have other things in the future.

IanEdington (approved):

grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
tdresser: Yup, good call.

WDYT about CreateBulkContactInDto? (or RequestDTO).

IanEdington: `CreateBulkContactRequestDTO` feels good to me. I wonder if `BulkCreateOut` should become `CreateBulkContactResponseDTO`.

tdresser: My thought was that BulkCreateOut might be agnostic to the type of the object bulk created, but in retrospect, that's probably silly. 

SGTM.

PR #53, opened 2025-05-15
Dockerize

tdresser:
This creates a new docker container for running the frontend and backend.

The database containers no longer have ports exposed on the host.

Note that the vscode dev container is still not 100%, but it's mostly working... Happy to leave it out of this PR, or fix it before landing.

Does this general direction seem right to you? This patch does mean that running things outside of docker is a bit of a headache - you need to expose some extra ports, and tweak some environment files.

We probably only want to support one dev approach - within docker or not within docker.

Does IntellliJ give you a nice way to develop within a container? You could maybe ssh into the grassroots_dev container, if nothing else?


IanEdington:
It looks like Intellij does let you develop in a dev container! https://www.jetbrains.com/help/idea/connect-to-devcontainer.html

IanEdington (approved):
Looks awesome. LGTM!

IanEdington:
I think the backend PR is failing the Backend Tests because we're using the docker compose file. Do you want me to push a fix to this PR?

resending to test the slack integration

IanEdington (approved):
LGTM! (once the tests are fixed)

PR #54, opened 2025-05-16
Add caddy as a reverse proxy

tdresser:
For now, we just use http not https.

This will let us serve the frontend and backend from the same origin, allowing us to use cookies for auth etc, and not worry about CORS.

Note the additional instruction in the README.md:

Add

```
$LOCAL_IP grassroots.local
```

to your /etc/hosts.

That's:
```
127.0.0.1 grassroots.local
```
if you're developing on the machine you're browsing from.


IanEdington (approved):
LGTM!

PR #55, opened 2025-05-18
feat(dev): install child packages when root installs

IanEdington:
automatically install child projects when the root project is installed


tdresser (approved):
LGTM, though this may need to change depending on how we containerize.

Issue #60, opened 2025-05-20
Investigate cleaning up Route param type conversion


tdresser:
If possible, would it make sense to do this (map empty string to undefined) across all routes for all query params?

_Originally posted by @tdresser in https://github.com/gpo/grassroots/pull/52#discussion_r2091287335_

PR #61, opened 2025-05-20
Get rid of implicit type conversions

tdresser:
The class-transformer [docs](https://github.com/typestack/class-transformer?tab=readme-ov-file#implicit-type-conversion) on implicit conversion say: 

    NOTE If you use class-validator together with class-transformer 
    you propably DON'T want to enable this function.

There's apparently a bunch of bad behavior where "false" will end up true, and stuff like that.

Instead, we should define explicit mechanisms for converting between types where appropriate. This lets us centralize specific special transforms we want (e.g., an id of "" should be treated as undefined, as in https://github.com/gpo/grassroots/issues/60).

### DIFF
--- a/grassroots-backend/src/contacts/Contacts.service.ts
+++ b/grassroots-backend/src/contacts/Contacts.service.ts
@@ -54,7 +54,7 @@ export class ContactsService {
         lastName: LikeOrUndefined(contact.lastName),
         email: LikeOrUndefined(contact.email),
         phoneNumber: LikeOrUndefined(contact.phoneNumber),
-        id: contact.id ? Equal(contact.id) : undefined,
+        id: contact.id !== undefined ? Equal(contact.id) : undefined,
       },
     });
     return {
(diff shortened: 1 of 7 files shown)


IanEdington (approved):
If I'm reading the docs right, class-transformer enforces stricter type safety by default by not allowing implicit conversions, which makes sense to me. And the `@Transform` is how we make an explicit conversion.

LGTM

tdresser:
The class-transformer docs (https://github.com/typestack/class-transformer?tab=readme-ov-file#implicit-type-conversion) on implicit conversion say:

"NOTE If you use class-validator together with class-transformer 
you propably DON'T want to enable this function."

There's apparently a bunch of bad behavior where "false" will end up true, and stuff like that.

Instead, we should define explicit mechanisms for converting between types where appropriate. This lets us centralize specific special transforms we want (e.g., an id of "" should be treated as undefined, as in https://github.com/gpo/grassroots/issues/60).

PR #64, opened 2025-05-21
fix: missing ValidateNested

IanEdington:
Validation isn't recursive by default:
https://github.com/typestack/class-validator#validating-nested-objects

We are investigating if an eslint rule exists


tdresser (approved):
Thanks!

PR #70, opened 2025-05-26
Migrate from TypeORM to MikroORM.

tdresser:
Most of my time here was spent addressing some issues which TypeORM papered over, in particular, it happily accepted a string when querying for a number.

I could pull that fix out into a separate CL if you want.


grassroots-backend/src/mikro-orm.config.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable check-file/filename-naming-convention */
```
tdresser, line 1: This is required for the mikro orm CLI to work, for running migrations and stuff. A bit annoying that we need to read the env file in two places, but I think it's probably fine.

grassroots-frontend/src/routes/Search.tsx
```diff
@@ -19,14 +20,18 @@ const ROWS_PER_PAGE = 10;
...
+      { mode: "sync" },
+    ),
     mode: "onChange",
   });
 
   const [rowsToSkip, setRowsToSkip] = useState<number>(0);
 
   const searchParams: PaginatedContactSearchInDTO = {
-    contact: form.watch(),
+    contact: plainToInstance(ContactSearchInDTO, form.watch()),
```
tdresser, line 34: This definitely shouldn't have worked, but TypeORM happily took a string instead of a number...

IanEdington (commented):
😅 WOW that was a marathon. I don't think you need to break out the type fixes between TypeORM and MikroORM.

If you merge this PR now, it will be merged into the `auth` branch rather than the `main` branch. I'll hold off giving this a stamp until auth is merged since I've seen people accidentally merge SO many times.

grassroots-backend/src/testing/E2eSetup.ts
```diff
@@ -28,25 +26,29 @@ export function useE2ETestFixture(
...
     const { port } = await listenAndConfigureApp(app, 0);
     const grassrootsAPI = createClient<paths>({
       baseUrl: `http://localhost:${String(port)}`,
       credentials: "include",
     });
-    fixture = new E2ETestFixture({ app, grassrootsAPI, queryRunner });
+    const orm = app.get<MikroORM>(MikroORM);
+    const generator = orm.getSchemaGenerator();
+    await generator.dropSchema();
+    await generator.createSchema();
```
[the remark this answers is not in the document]
tdresser: We need to make sure the relevant tables exist somehow. That could alternatively be running migrations on the test database as well as the dev database. Or maybe there's a global "before anything runs" jest hook?

Aha! This should move to: 
https://jestjs.io/docs/configuration#globalsetup-string

IanEdington (approved):

PR #71, opened 2025-05-26
Add eslint rules for nest and strict boolean checks.

tdresser:


grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
```diff
@@ -75,23 +78,34 @@ export class ContactSearchInDTO {
   @IsInt()
   @Min(1)
   id?: number;
+  @IsOptional()
   email?: string;
+  @IsOptional()
   firstName?: string;
+  @IsOptional()
   lastName?: string;
+  @IsOptional()
```
IanEdington, line 87: Is this required in NestJS if there's a `?`?

grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
```diff
@@ -75,23 +78,34 @@ export class ContactSearchInDTO {
...
   phoneNumber?: string;
 }
 
 export class PaginatedContactSearchInDTO {
   @ValidateNested()
+  @Type(() => ContactSearchInDTO)
   contact!: ContactSearchInDTO;
+
   @ValidateNested()
+  @Type(() => PaginatedInDTO)
```
IanEdington, line 97: Is NestJS not able to infer this from `paginated!: PaginatedInDTO`?

grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
tdresser: Yeah. There are two ways Nest uses "reflection".

1. Reflect-metadata, which is just about inspecting decorators and their parameters.
2. The nest openAPI cli plug-in, which does a very limited set of things 

Neither of those handle this case. The nest eslint plug-in catches this, making it a bit less annoying.

grassroots-backend/src/grassroots-shared/Contact.entity.dto.ts
IanEdington: These both seem like issues we'd want linting for. So easy to miss, and it seems like they'd be pretty important.

tdresser: Yup, and eslint-plugin-nestjs-typed provides rules for both of these.

PR #72, opened 2025-05-26
docs: project vision (incomplete)

IanEdington:


MuhammadElsoukkary (commented):
eveything seems to be checked off in terms of reasons and goals maybe specify that is planned to be an open source web app and the features it will aim to have or how we plan to differ from others

PR #77, opened 2025-06-04
Migrate to vite for backend

tdresser:
Migrate to vite for the backend, and vitest for testing.

This gives us a consistent way of running / bundling for the frontend, and backend, and tests, whereas previously each of these were run with a different stack.

This merges the e2e and unit tests. We could undo this, but it did help me identify a flaky issue with the MikroORM rollback strategy, which this patch also fixes.

### DIFF
--- a/grassroots-backend/src/contacts/Contacts.service.ts
+++ b/grassroots-backend/src/contacts/Contacts.service.ts
@@ -5,43 +5,36 @@ import {
   CreateContactInDto,
   PaginatedContactOutDTO,
   PaginatedContactSearchInDTO,
-} from "../grassroots-shared/Contact.entity.dto";
-import { InjectRepository } from "@mikro-orm/nestjs";
-import { EntityRepository, FilterQuery } from "@mikro-orm/core";
+} from "../grassroots-shared/Contact.entity.dto.js";
+import { EntityManager, FilterQuery } from "@mikro-orm/core";
 import { LikeOrUndefined } from "../util/LikeOrUndefined";
 
 @Injectable()
 export class ContactsService {
-  constructor(
-    @InjectRepository(ContactEntityOutDTO)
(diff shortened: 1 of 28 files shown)


grassroots-backend/src/vite-env.d.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable check-file/filename-naming-convention */
```
tdresser, line 1: Ideally we might share this with the frontend, but I didn't see a nice way of doing so.

grassroots-backend/src/contacts/Contacts.service.ts
```diff
@@ -5,43 +5,36 @@ import {
...
 
 @Injectable()
 export class ContactsService {
-  constructor(
-    @InjectRepository(ContactEntityOutDTO)
-    private readonly contactsRepository: EntityRepository<ContactEntityOutDTO>,
-    // Will be needed for explicit flushing.
-    // private readonly em: EntityManager,
-  ) {}
+  constructor(private readonly entityManager: EntityManager) {}
```
IanEdington, line 14: This change and the above `useFactory` provider change above seem to be related. Both seem to be negative tradeoffs of moving to vitest (unless I'm miss reading). We used to be able to use `contactsRepository` which had type information about the entity it was interacting with, and we used to be able to inject `ContactsService` (and I'm assuming future Services) directly in tests.

Is this something that's worth documenting why this tradeoff exists and what the benefits are?

IanEdington: I see that in #79 you address the module part of this comment. LGTM and iterate after these two PRs are in.

IanEdington (approved):
This is a very clean PR. The migration to vitest is a lot nicer than I expected. Thanks for putting this together.

PR #79, opened 2025-06-07
Visualize dependency graph, and fix the issues this highlighted

tdresser:
Once I visualized our dependency graph, it became clear that my whole mental model of how dependencies worked was wrong. Nest seems to work best when you reason about dependencies at the module level, which I wasn't doing.

This patch visualizes the dependency graph (for dev and in 1 example test), and addresses a few issues highlighted by the visualization.


docs/DependencyGraphForTest.md
```diff
@@ -0,0 +1,9 @@
+```mermaid
```
IanEdington, line 1: 🤩 

I love mermaid. So handy

grassroots-backend/src/contacts/Contacts.service.ts
```diff
@@ -6,35 +6,36 @@ import {
...
+import { EntityManagerProvider } from "../orm/EntityManager.provider.js";
 
 @Injectable()
 export class ContactsService {
-  constructor(private readonly entityManager: EntityManager) {}
+  repo: EntityRepository<ContactEntityOutDTO>;
+  constructor(private readonly entityManagerProvider: EntityManagerProvider) {
+    this.repo =
+      entityManagerProvider.entityManager.getRepository<ContactEntityOutDTO>(
+        ContactEntityOutDTO,
```
IanEdington, line 19: Continuing from https://github.com/gpo/grassroots/pull/77#discussion_r2134815188

In a fully featured dependency injection framework, I would expect to be able to inject the `repo: EntityRepository<ContactEntityOutDTO>` here.

IanEdington (commented):
🙌 This is AMAZING! I love those dependency graphs! and I'm glad it lead to a better understanding of NestJS's DI tool.

Only not approving to avoid accidently squashing the stacked PRs

grassroots-backend/src/contacts/Contacts.service.ts
tdresser: Without care, the trouble is that if we inject the repo, the repo gets the global EntityManager instead of the one we control for testing. We could probably do it with a custom @InjectRepository decorator?

PR #80, opened 2025-06-07
Using cascading env files

tdresser:
Check in config that is public and shared by all developers.


setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 npm install
 
-cp .env.development.sample .env.development
+cp .env.development.local.sample .env.development.local
 cp .env.test.sample .env.test
```
IanEdington, line 5: What's the plan for test? Can we checking .env.test and not use .env.test.local? We could remove this line and the .env.local.sample.

This would only work if config for ci test and local testing is checked in separately, which I believe it is.

IanEdington (approved):

setup.sh
tdresser: I think what you're describing is what this patch does, except for removing this line. 😄 

I think we actually don't need this step at all, since .env.development should contain everything we need for tests.

Removing this step, and trying again.

PR #85, opened 2025-06-11
Integrate Google Signin into backend

tdresser:


.env.development
```diff
@@ -3,4 +3,6 @@ POSTGRES_PASSWORD = localDevPassword
 POSTGRES_HOST = db
 POSTGRES_DATABASE = grassroots
 POSTGRES_PORT = 5432
-VITE_BACKEND_HOST = http://grassroots.local/api
+VITE_BACKEND_HOST = http://grassroots.org/api
```
tdresser, line 6: Switching to .org was required to get Google signin to work.

grassroots-backend/package.json
```diff
@@ -53,6 +54,7 @@
     "@types/node": "^22.15.3",
     "@types/passport": "^1.0.17",
     "@types/passport-local": "^1.0.38",
+    "@types/passport-openidconnect": "^0.1.3",
```
tdresser, line 57: passport-google-oicd is untyped, but the types are mostly from here, so I just used these.

grassroots-backend/src/App.controller.ts
```diff
@@ -1,40 +1,13 @@
...
-    private authService: AuthService,
-  ) {}
+  constructor(private readonly appService: AppService) {}
 
   @Get()
   getHello(): HelloOutDTO {
     return this.appService.getHello();
   }
-
-  @UseGuards(DefaultAuthGuard)
```
tdresser, line 22: This should have been in the auth controller before.

grassroots-backend/src/auth/Auth.controller.ts
```diff
@@ -1,32 +1,76 @@
...
+  @Get("is_authenticated")
+  isUserLoggedIn(@Request() req: GrassrootsRequest): LoginStateDTO {
+    return { isLoggedIn: req.isAuthenticated(), user: req.user };
+  }
+
+  // This is an example of using user info, to enable a test.
+  // TODO: remove this once we have real routes using user info.
+  @Get("example_route_using_user")
+  @UseGuards(DefaultAuthGuard)
+  // Not sure why UseGuards breaks the OpenAPI plugin.
```
tdresser, line 59: :(

grassroots-backend/src/auth/GoogleOAuth.strategy.ts
```diff
@@ -1,27 +1,73 @@
...
-  async validate(email: string, password: string): Promise<UserEntity | null> {
-    const user = await this.authService.validateUser(email, password);
-    if (!user) {
-      throw new UnauthorizedException();
+  validate: VerifyFunction = async (
+    issuer: string,
+    profile: Profile,
+    done: VerifyCallback,
+  ): Promise<void> => {
+    const email = profile.emails?.[0];
```
tdresser, line 50: Just using the first email address feels a bit weird, but is probably fine?

We could use a unique ID from Google instead of email as a unique id.

grassroots-backend/test/Auth.e2e-spec.ts
```diff
@@ -1,25 +1,49 @@
-import { useE2ETestFixture } from "../src/testing/E2eSetup";
 import { describe, expect, it } from "vitest";
-import { AppModule } from "../src/App.module";
+import { useE2ETestFixture } from "../src/testing/E2eSetup";
```
tdresser, line 2: I probably spent more time trying to get tests to work than getting the functionality working.

This isn't great, but nailing this stuff is pretty subtle, and mostly just testing 3p libraries. 

Ideally we'd have a test that this does actually insert a user into the database. Happy to poke at it more if you think that's worthwhile, but I wanted to at least get this in front of you in the current state.

tdresser (commented):
Testing of this isn't great, but getting tests right is pretty subtle, and mostly just testing 3p libraries. 

Ideally we'd have a test that this does actually insert a user into the database. Happy to poke at it more if you think that's worthwhile, but I wanted to at least get this in front of you in the current state.

EDIT: added a unit test that mostly does this!

grassroots-backend/src/auth/GoogleOAuth.strategy.ts
IanEdington: `unique ID from Google` seems like a safer option. Is there such an ID that we're provided from google?
I imagine the case where the primary email address changes or where the google account someone signs up with doesn't have the same email address they were invited with (invite sean@hotmail.com, signup with sean.smith@gmail.com).

IanEdington (approved):
Looks great!

PR #86, opened 2025-06-12
Removed eclint

markwong167:
Removed eclint from the code base.


IanEdington (dismissed):
based on our converstion in slack we're going ahead with removing even though we're loosing some functionality since the package is causing issues on windows and doesn't add a lot of value

PR #91, opened 2025-06-17
Infra for custom eslint rules 

tdresser:
This implements a rule that ensures any class ending in "DTO" only has properties which are definite or optional (end in ! or ?).

This rule doesn't make a ton of sense in isolation, it needs a companion rule which enforces that these classes have no constructors, but I'll land that in a followup.

This isn't integrated as a monorepo, partially because we haven't yet figured out our broader monorepo strategy, and partially because the way eslint depends on things is a bit different from normal typescript monorepo dependencies, so it'll take some figuring out.


eslint_rules/src/rules/dto-and-entity-style.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      tsconfigRootDir: path.join(__dirname, "../.."),
+    },
+  },
+});
+
+ruleTester.run("no-loop-over-enum", rule, {
+  valid: [
+    `class FooDTO { a!: number; b?: number}`,
+    `class FooDTO { a!: number}`,
+    `class FooDTO { a?: number}`,
```
IanEdington, line 30: Could we throw in a class that doesn't end with DTO like `class FooBar { a: number}`

IanEdington (approved):
🙌

PR #98, opened 2025-06-18
Creates a new symlink each time

markwong167:
Will be replaced later with docker bind mount, but this would make coding actually doable for those on Windows


tdresser (approved):
LGTM!

PR #99, opened 2025-06-18
Land config for git-town

tdresser:
I'm using git-town for stacked changes, and it seems pretty good.

Okay to land the config in-repo? Or should I add it to .gitignore?


IanEdington (approved):
I'd probably lean towards using `.gitignore` on `.git-branches.toml` unless we all want to align on using git-town, but it's a weak opinion

tdresser:
Done.

PR #106, opened 2025-06-21
Implement backend route auth

tdresser:
This splits the Auth Guard in two, with one guard for performing OAuth redirects, and another for validating the session for API calls.

It's a bit clunky getting the two guards to collaborate, we use a decorator read by the Session guard to turn off the Session Guard for routes where we want the OAuth redirect guard.

The frontend doesn't respond well to 401 responses from the backend at this point.


grassroots-backend/src/testing/GetTestApp.ts
```diff
@@ -52,7 +52,7 @@ export async function getTestApp(
   });
   builder = overrideEntityManagerForTest(builder);
   if (dependencies.overrideAuthGuard === true) {
-    builder = builder.overrideGuard(DefaultAuthGuard).useClass(MockAuthGuard);
+    builder = builder.overrideProvider(SessionGuard).useClass(MockAuthGuard);
```
IanEdington, line 55: MockSessionGuard?

grassroots-frontend/src/GrassRootsAPI.ts
```diff
@@ -3,4 +3,5 @@ import { paths } from "./grassroots-shared/OpenAPI.gen";
 
 export const grassrootsAPI = createFetchClient<paths>({
   baseUrl: import.meta.env.VITE_BACKEND_HOST,
+  //requestInitExt: { credentials: "same-origin" },
```
IanEdington, line 6: ?

grassroots-frontend/src/routes/CreateContact.tsx
```diff
@@ -25,6 +25,7 @@ function CreateContact(): JSX.Element {
     mutationFn: async (contact: CreateContactInDto) => {
       const result = await grassrootsAPI.POST("/contacts", {
         body: contact,
+        credentials: "include",
```
IanEdington, line 28: Will we need to do this on every route?

grassroots-backend/src/auth/Session.guard.ts
```diff
@@ -0,0 +1,38 @@
...
+export class SessionGuard {
+  constructor(private reflector: Reflector) {}
+
+  canActivate(context: ExecutionContext): boolean {
+    const isRoutePublic = this.reflector.getAllAndOverride<boolean>(
+      DECORATOR_METADATA_INDICATING_PUBLIC_ROUTE,
+      [context.getHandler(), context.getClass()],
+    );
+
+    const isOAuthRoute = this.reflector.getAllAndOverride<boolean>(
```
IanEdington, line 21: Is there a reason we need this to be distinct from a public route? It seems like the same logic could be achieved using `@PublicRoute`

grassroots-backend/openAPI.json
```diff
@@ -32,37 +32,30 @@
         ]
       }
     },
-    "/auth/login": {
-      "get": {
-        "operationId": "AuthController_login",
+    "/contacts": {
```
IanEdington, line 35: There's a pretty big diff here. I would have expected openAPI's ordering to be stable. I wonder if we should add a sort to this object.

IanEdington (approved):

grassroots-frontend/src/GrassRootsAPI.ts
tdresser: Removed.

grassroots-frontend/src/routes/CreateContact.tsx
tdresser: We actually don't need it on any route: cookies are included in same-origin requests by default. Removed.

grassroots-backend/src/auth/Session.guard.ts
tdresser: Done.

grassroots-backend/openAPI.json
tdresser: I'll address this in my formatting PR.

PR #108, opened 2025-06-22
Delete metadata.ts

IanEdington:
When the repository is first initialized (initial download or in CI), the metadata.ts doesn't exist. 
This swaps in the formatted metadata for metadata.ts when it doesn't exist.


IanEdington:
@tdresser This is one I think you should merge into your branch

PR #109, opened 2025-06-22
dev: making devcontainers easier

IanEdington:
I've been using devcontainers in vscode exclusively since Wednesday, and I'm not hating it.

This is what I've found, keep in mind I'm coming from Intellij so I'm not sure all my suggestions make sense

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,7 +1,14 @@
 {
   "customizations": {
     "vscode": {
-      "extensions": ["esbenp.prettier-vscode", "dbaeumer.vscode-eslint"]
+      "extensions": [
+        "esbenp.prettier-vscode",
+        "dbaeumer.vscode-eslint",
+        "bierner.markdown-mermaid",
+        "ms-azuretools.vscode-docker",
+        "github.vscode-github-actions",
+        "editorconfig.editorconfig"
+      ]
     }
   },
   "dockerComposeFile": ["docker/compose.yaml"],
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,8 @@
 #!/bin/bash
+
+if [ -z "$IN_DEVCONTAINER" ]; then
+  echo "Not running in the devcontainer. Skipping pre-commit hook."
+  exit 0
+fi
+
 npx lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# Only run pre-commit checks if we're running in the devcontainer
+# This is a temporary workaround until #100 is merged
 if [ -z "$IN_DEVCONTAINER" ]; then
   echo "Not running in the devcontainer. Skipping pre-commit hook."
   exit 0
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,16 +6,10 @@
       "request": "launch",
       "name": "NestJS gen-files-only",
       "program": "${workspaceFolder}/grassroots-backend/node_modules/.bin/nest",
-      "args": [
-        "start",
-        "--",
-        "--gen-files-only"
-      ],
+      "args": ["start", "--", "--gen-files-only"],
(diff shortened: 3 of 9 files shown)


.devcontainer.json
```diff
@@ -1,7 +1,14 @@
...
   "customizations": {
     "vscode": {
-      "extensions": ["esbenp.prettier-vscode", "dbaeumer.vscode-eslint"]
+      "extensions": [
+        "esbenp.prettier-vscode",
+        "dbaeumer.vscode-eslint",
+        "bierner.markdown-mermaid",
+        "ms-azuretools.vscode-docker",
+        "github.vscode-github-actions",
+        "editorconfig.editorconfig"
```
IanEdington, line 10: These are all the suggested plugins I kept getting asked to install

.husky/pre-commit
```diff
@@ -1,2 +1,8 @@
 #!/bin/bash
+
+if [ -z "$IN_DEVCONTAINER" ]; then
```
IanEdington, line 3:
```suggestion
# Only run pre-commit checks if we're running in the devcontainer
# This is a temporary workaround until #100 is merged
if [ -z "$IN_DEVCONTAINER" ]; then
```

.husky/pre-commit
```diff
@@ -1,2 +1,8 @@
 #!/bin/bash
+
```
IanEdington, line 2:
```suggestion

# Only run pre-commit checks if we're running in the devcontainer
# This is a temporary workaround until #100 is merged
```

grassroots-backend/package.json
```diff
@@ -7,9 +7,7 @@
   "license": "UNLICENSED",
   "scripts": {
     "build": "nest build",
-    "start": "nest start",
-    "start:dev": "nest start --watch",
-    "start:debug": "nest start --debug --watch",
+    "start": "nest start --watch --debug",
```
IanEdington, line 10: We talked about having `npm start` be the "perfect" dev command. I know that's not actually possible but this is my attempt.

IanEdington: Actually... we didn't. It was on the agenda, but not discussed.

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "NestJS gen-files-only",
```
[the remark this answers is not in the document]
tdresser: I currently don't use the vscode launch stuff. 

Maybe I should though? No real opinions on my end. I don't mind this existing, but I'm unlikely to use it.

.devcontainer.json
tdresser: Thanks!

grassroots-backend/package.json
tdresser: LGTM

I wonder if the `test` commands should default to watching as well?

That would require splitting the tstyche tests into a separate command, but might still be worth it.

tdresser (approved):
LGTM, with a few optional ideas.

grassroots-backend/package.json
IanEdington: done

.vscode/launch.json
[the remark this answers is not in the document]
MuhammadElsoukkary: I can give it a try to I dont think I use it maybe I do

Dockerfile
```diff
@@ -18,6 +18,11 @@ RUN deluser node --remove-home \
     && groupadd -g ${GID} ${UNAME} \
     && useradd -g ${UNAME} -u ${UID} ${UNAME} -m
 
+# Install additional tools
+RUN apt-get update \
+    && apt-get install -y zsh \
```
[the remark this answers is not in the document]
IanEdington: I'll try that approach.

My thinking was that doing it in the image would mean it doesn't have to be re-run every time the containers need to be rebuilt. I'm happy to try this approach first though, and see if I notice the startup time.

PR #113, opened 2025-06-25
Entity and DTO cleanup

tdresser:
Keep entities as backend only concepts. This means we can get rid of the mikro orm shim and have a much clearer conceptual boundary between entities and DTOs, at the cost of a bit of duplication. 

We do need to make sure we keep the DTOs and entities in sync, though you can see an example in ContactEntity of enforcing this at least to some degree.

I also renamed all "InDTO"s to "RequestDTO" and "OutDTO"s to "ResponseDTO".


IanEdington (commented):
I like the general idea of this PR, but I'm worried about how much duplication this will cause.

I feel like there are two types of duplication with entities / DTOs. One is code duplication, which doesn't bother me. The other is cognitive(?) duplication, having to hold multiple versions of the same object in our minds as we're working. I really don't want to add cognitive duplication if at all possible. I believe that having less strong types on the objects is better than having multiple conceptual versions of the same object.

Some ideas for how to solve this:
- I wonder if there's a way to generate one type from the other?
- CI linting rule to make sure they don't drift significantly?
- I think having a separate Requests and Responses for each route still makes sense. Could we reduce the number of Request and Response DTOs by only having them at the root level? For example, can we merge `CreateContactRequestDto` and `ContactResponseDTO` by making `CreateContactRequestDto` a partial of `ContactResponseDTO`?

I don't mind the shims. Is there a reason we should remove them?

grassroots-backend/src/grassroots-shared/Contact.dto.ts
```diff
@@ -78,37 +62,34 @@ export class ContactSearchInDTO {
...
-  lastName?: string;
-  @IsOptional()
-  phoneNumber?: string;
 }
 
-export class PaginatedContactSearchInDTO {
+export class ContactSearchRequestDTO extends IntersectionType(
+  OmitType(PartialType(ContactDTO), ["id"]),
+  ContactSearchRequestId,
+) {}
```
IanEdington, line 70: My understanding is that we need the decorators on these classes for validation. With all the type modifiers like `IntersectionType`, do the decorators exist on the resulting instance?

grassroots-backend/src/users/User.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+  firstName?: string;
+  @Property({ nullable: true })
+  lastName?: string;
+  @Property({ nullable: true })
+  displayName?: string;
+
+  foo!: number;
+}
+
+export type Check = AssertPropsEqual<UserDTO, UserEntity>;
```
IanEdington, line 22: Do we need to export this type because otherwise it's seen as useless?

grassroots-backend/src/grassroots-shared/util/TypesEqual.ts
```diff
@@ -0,0 +1,8 @@
+import { PropsOf } from "./PropsOf";
```
IanEdington, line 1: would it make sense to update the filename to `AssertPropsEqual`?

IanEdington (approved):
I'm loving this version!

grassroots-backend/src/grassroots-shared/util/TypesEqual.ts
tdresser: Done.

PR #128, opened 2025-07-01
update @swc/core and @swc/cli

tprost:
on MacOS I had to reinstall @swc/core and @swc/cli to get the backend running


grassroots-backend/package.json
```diff
@@ -48,7 +48,7 @@
     "@nestjs/schematics": "^11.0.0",
     "@nestjs/testing": "^11.0.1",
     "@swc/cli": "^0.6.0",
```
tdresser, line 50: Looks like you're only updating core?

tdresser (approved):

PR #133, opened 2025-07-03
dev: use https with self signed cert

IanEdington:


docker/caddy/Caddyfile
```diff
@@ -9,3 +10,7 @@ http://grassroots.org {
         reverse_proxy grassroots_dev:5173
     }
 }
+
+http://grassroots.org {
+    redir https://grassroots.org{uri} permanent
+}
```
IanEdington, line 16:
```suggestion
}

```

IanEdington:
> Could you add instructions on installing the cert to the readme?

sorry about that. I re-read my comment and realized it's not very clear. I didn't work on getting the self-signed cert into the browser, which is why I don't think it's a complete solution. I'll add notes to #104

PR #135, opened 2025-07-04
Implement basic Role entity - hardcoded edition

tdresser:


tdresser:
"Another option could be to have the roles in the database and update them on server start. This feels less clean, but there would be a clear many-to-many relationship."

That's roughly the approach I took [here](https://github.com/gpo/grassroots/blob/f262925b05b94dd23cd0e1ec5a3191382d78c800/grassroots-backend/src/organizations/Roles.service.ts) in [this PR](https://github.com/gpo/grassroots/pull/123), right?

Because we need metadata on the [join table](https://docs.google.com/document/d/1HaCu3ZmWBbfXSs430Q3Ojb6NGwEecdb5Il2nhtZadqI/edit?tab=t.da2vnvo60x86#bookmark=id.19et7cxai2xr), and not just a straight mapping from ID to ID, we need an explit "UserRole" entity either way. mikroorm doesn't support custom join tables, so the way to model this is a one to many relationship between User -> UserRole and a one-to-one relationship between UserRole and Role.

"We could just map it to the fixed IDs of the Roles"
That's how I was imagining it for now. I don't have a strong preference either way. I don't think it'll be hard to model in the orm layer because we need an explicit UserRole entity regardless.

PR #137, opened 2025-07-05
added migration check

markwong167:
Added a migration check to github actions.
If there are migrations that need to be created, it should fail.


tdresser:
I don't care about the branch name at all.

tdresser (approved):
LGTM, but make the change you suggested, and remove the change causing the test to fail.

PR #138, opened 2025-07-06
Don't disable the submit button on Create Contact form

tprost:


tdresser (approved):
That's a simpler solution than I anticipated! 

Thanks!

PR #143, opened 2025-07-06
Add button for adding random data

tdresser:
This is a replacement for https://github.com/gpo/grassroots/pull/19 .

As we chatted about a while back, I like having this kind of development tool in the frontend, as it lets us test some of the caching / cache busting behavior built into tanstack query, which we're likely to rely on for offline support. This is nicer if the button is in the sidebar, not on a separate route.

I'm planning to include "Add a tree of organizations" to the functionality of this button at some point (and change it's title).

### DIFF
--- a/grassroots-frontend/package-lock.json
+++ b/grassroots-frontend/package-lock.json
@@ -8,6 +8,7 @@
       "name": "grassroots-frontend",
       "version": "0.0.0",
       "dependencies": {
+        "@faker-js/faker": "^9.9.0",
         "@mantine/core": "^7.17.4",
         "@mantine/hooks": "^7.17.4",
         "@tanstack/react-query": "^5.74.11",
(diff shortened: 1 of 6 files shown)


tprost:
Not a big fan of this, but if you're gonna keep it, make sure it's labelled "(for testing)" so newcomers like me don't get confused.

Hopefully I can put together some nice e2e tests that convince you you don't need that button :upside_down_face:

tdresser:
I'll put a bit of work into making it clearer what this is before landing.

IanEdington:
Something I've used in the past that works well is a hidden "developer tools" component that's opened using a keyboard shortcut, or a special swipe gesture on mobile.

In the future, we could gate it behind some kind of security, and completely remove parts from prod (aka data gen).

The benefit I see to this approach is that it can help with debugging at multiple stages, not just local development and can be made to work on all device types.

Have either of you worked with tools like this? How have you found them?

tdresser:
We already have a link for the tanstack router dev tools, which kinda fits in this camp.

Chrome is jam packed with this kind of thing, and it's pretty critical in that context. I don't have recent professional frontend web dev experience though.

PR #145, opened 2025-07-07
Implement basic Role entity - hardcoded edition

tdresser:
Continuing https://github.com/gpo/grassroots/pull/135 ...

I have no idea why git town closed that pull request... Seems like it works perfectly except when it doesn't...


grassroots-backend/src/organizations/Roles.service.ts
```diff
@@ -0,0 +1,40 @@
...
+      Permission.VIEW_CONTACTS,
+      Permission.MANAGE_CONTACTS,
+      Permission.MANAGE_USERS,
+    ],
+  },
+];
+
+@Injectable()
+export class RolesService {
+  findAll(): RoleEntity[] {
```
IanEdington, line 37: Should we normally return entities or DTOs from services? Seems like we would want a convention on this

IanEdington (approved):

grassroots-backend/src/migrations/Migration20250630182244.ts
```diff
@@ -0,0 +1,13 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20250630182244 extends Migration {
```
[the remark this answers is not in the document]
tdresser: Oops, thanks!

grassroots-backend/src/organizations/Roles.service.ts
tdresser: Switched to DTOs.

PR #146, opened 2025-07-07
Run Mikroorm Migration on backend poststartup

markwong167:


grassroots-backend/src/Main.ts
```diff
@@ -68,6 +69,12 @@ async function writeFormattedMetadata(): Promise<void> {
   });
 }
 
+async function createMikroORMMigration(): Promise<void> {
+  const orm = await MikroORM.init();
```
tdresser, line 73: You might be able to get this from `app`. I think this is dependency injected, and anything injected you can pull out via app.get. (not positive, but worth a shot) 

Otherwise, I think we need to close this, right? I think every init is supposed to have a matching close?

tdresser (commented):
Have you tested what happens if you get an entity really out of sync, and then run this? 

E.g., rename one of the fields in Contact Entity?

It's possible that this will cause nest to throw during app init, in which case this approach won't work and we'll need to run this before app init.

grassroots-backend/src/Main.ts
markwong167: That's true, no wonder.

markwong167:
> Have you tested what happens if you get an entity really out of sync, and then run this?
> 
> E.g., rename one of the fields in Contact Entity?
> 
> It's possible that this will cause nest to throw during app init, in which case this approach won't work and we'll need to run this before app init.

It looks like everything still runs after renaming it and getting the error.
image

grassroots-backend/src/Main.ts
```diff
@@ -68,7 +70,29 @@ async function writeFormattedMetadata(): Promise<void> {
...
+    console.error(e);
+  } finally {
+    await orm.close(true);
+  }
+}
+
 async function bootstrap(port: number): Promise<void> {
+  console.time("pre-startup");
+  const preStartupTasks = [createMikroORMMigration()];
+  console.timeEnd("pre-startup");
```
tdresser, line 94: Time end should go after we await the promises.

grassroots-backend/src/Main.ts
```diff
@@ -68,7 +70,29 @@ async function writeFormattedMetadata(): Promise<void> {
   });
 }
 
+async function createMikroORMMigration(): Promise<void> {
+  const orm = await MikroORM.init(mikroORMConfig);
+  const migrator = orm.getMigrator();
+  try {
+    await migrator.createMigration().then((migrationResult) => {
```
tdresser, line 77: Does this run the migration, or just create it?

markwong167: Sorry about that. Thanks for reviewing. I should have given my code a better look.
I think that I was craving the texan bbq when I pushed it that my mind went blank.

tdresser: No worries at all! Mmmm barbecue does sound good...

tdresser (commented):
Where did that migration come from? It doesn't look like it should exist...

Issue #90, opened 2025-06-15
Build our own ConfigModule


tdresser:
Properties we'd like to have in a ConfigModule:
- useable outside of nestjs, for the mikroorm-setup script.
- stateless, so it doesn't muck around with process env vars, but just pulls values out of env files and provides access to them.
- typed (you can do this with ConfigModule, but it's a bit more convoluted).

I think this is pretty straightforward using dotenv.parse.

tdresser:
Proposed steps:

1. Replace usage of dotenv-flow in mikro-orm.config.ts with dotenv.parse. You'll need to call dotenv.parse on all files in getEnvFilePaths and combine their results. (This can be done using the javascript spread operator).

2. Factor this out into a class (probably not yet named ConfigModule to avoid collisions), with one property per environment variable. Use class-validator decorators to validate the properties. I think a static "Create" method which creates an instance of this and uses the `cast` method to create an instance and invoke the class-validator decorators would make sense.

3. Make this class @Injectable, and use it in place of ConfigModule throughout the codebase.

Sound reasonable?

MuhammadElsoukkary:
Yes it does

PR #150, opened 2025-07-09
Replace dotenv-flow with dotenv.parse in mikro-orm.config

MuhammadElsoukkary:
## Summary
Replace usage of `dotenv-flow` with explicit `dotenv.parse()` in the MikroORM configuration as the first step toward creating a unified, stateless configuration system.

**Related Issue:** #90

## Changes Made
- **Removed** `dotenv-flow` dependency and automatic process.env loading
- **Added** manual environment file parsing using `dotenv.parse()`
- **Maintained** same file priority logic with `.reverse()` (earlier files take priority)
- **Replaced** `process.env` usage with explicit `envConfig` object
- **Preserved** all existing functionality and behavior

## Technical Details
- Uses `getEnvFilePaths()` to get list of environment files
- Combines multiple `.env` files using spread operator
- No changes to global state (`process.env` is not modified)
- Maintains compatibility with existing environment file structure

## Testing
- ✅ Backend starts successfully in container
- ✅ MikroORM connects to database correctly
- ✅ All entities discovered and loaded
- ✅ Application routes mapped successfully
- ✅ Passes ESLint and Prettier checks

## Benefits
- **Explicit control** over environment variable loading
- **No global state pollution** (stateless approach)
- **Same behavior** as before but more predictable
- **Foundation** for upcoming AppConfig validation system

## Implementation Notes
This implements step 1 of the plan outlined in #90:
> "Replace usage of dotenv-flow in mikro-orm.config.ts with dotenv.parse. You'll need to call dotenv.parse on all files in getEnvFilePaths and combine their results."

## Next Steps
- Step 2: Factor out into validated class with class-validator decorators
- Step 3: Make class @Injectable and replace ConfigModule usage throughout codebase

### DIFF
--- a/grassroots-backend/src/mikro-orm.config.ts
+++ b/grassroots-backend/src/mikro-orm.config.ts
@@ -2,23 +2,46 @@
 import { defineConfig, PostgreSqlDriver } from "@mikro-orm/postgresql";
 import { UserEntity } from "./users/User.entity";
 import { ContactEntity } from "./contacts/entities/Contact.entity";
-
-import dotenvFlow from "dotenv-flow";
-import { getEnvFilePaths } from "./GetEnvFilePaths";
 import { OrganizationEntity } from "./organizations/Organization.entity";
-dotenvFlow.config({
-  // Reversed so that earlier files take priority, to align with the ConfigModule.
-  files: getEnvFilePaths().reverse(),
-});
+import * as dotenv from "dotenv";
+import { readFileSync, existsSync } from "fs";
+import { getEnvFilePaths } from "./GetEnvFilePaths";
+
+/**
+ * Loads and combines environment variables from multiple .env files
+ * Earlier files in the list take priority (same as dotenv-flow behavior)
+ */
+function loadEnvironmentVariables(): Record<string, string> {
+  // Reversed so that earlier files take priority, to align with the ConfigModule
+  const envFiles = getEnvFilePaths().reverse();
+  let combinedEnv: Record<string, string> = {};
+
+  for (const filePath of envFiles) {
+    if (existsSync(filePath)) {
+      try {
+        const fileContent = readFileSync(filePath, "utf8");
+        const parsedEnv = dotenv.parse(fileContent);
+        combinedEnv = { ...combinedEnv, ...parsedEnv };
+      } catch (error) {
+        console.warn(`Warning: Could not read env file ${filePath}:`, error);
+      }
+    }
+  }
+
+  return combinedEnv;
+}
+
+// Load environment configuration
+const envConfig = loadEnvironmentVariables();
 
 export default defineConfig({
   metadataCache: { enabled: false },
   driver: PostgreSqlDriver,
   entities: [ContactEntity, UserEntity, OrganizationEntity],
-  host: process.env.POSTGRES_HOST,
-  port: Number(process.env.POSTGRES_PORT),
-  user: process.env.POSTGRES_USER,
-  password: process.env.POSTGRES_PASSWORD,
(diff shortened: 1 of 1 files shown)


grassroots-backend/src/mikro-orm.config.ts
```diff
@@ -2,23 +2,46 @@
...
-
-import dotenvFlow from "dotenv-flow";
-import { getEnvFilePaths } from "./GetEnvFilePaths";
 import { OrganizationEntity } from "./organizations/Organization.entity";
-dotenvFlow.config({
-  // Reversed so that earlier files take priority, to align with the ConfigModule.
-  files: getEnvFilePaths().reverse(),
-});
+import * as dotenv from "dotenv";
+import { readFileSync, existsSync } from "fs";
```
tdresser, line 7: Nit: we should generally prefer fs/promises, as in some cases it'll allow parallel io.

In fact, we should be able to read all env files in parallel, using const contents = Promise.all(inputFilePromises)

grassroots-backend/src/mikro-orm.config.ts
```diff
@@ -2,23 +2,46 @@
...
+  let combinedEnv: Record<string, string> = {};
+
+  for (const filePath of envFiles) {
+    if (existsSync(filePath)) {
+      try {
+        const fileContent = readFileSync(filePath, "utf8");
+        const parsedEnv = dotenv.parse(fileContent);
+        combinedEnv = { ...combinedEnv, ...parsedEnv };
+      } catch (error) {
+        console.warn(`Warning: Could not read env file ${filePath}:`, error);
```
tdresser, line 26: I think we should probably throw in this case. If there an env file there but it can't be parsed, that implies it's actually broken somehow.

grassroots-backend/src/mikro-orm.config.ts
```diff
@@ -2,23 +2,46 @@
...
+  // Reversed so that earlier files take priority, to align with the ConfigModule
+  const envFiles = getEnvFilePaths().reverse();
+  let combinedEnv: Record<string, string> = {};
+
+  for (const filePath of envFiles) {
+    if (existsSync(filePath)) {
+      try {
+        const fileContent = readFileSync(filePath, "utf8");
+        const parsedEnv = dotenv.parse(fileContent);
+        combinedEnv = { ...combinedEnv, ...parsedEnv };
```
tdresser, line 24: If we flip this, we should be able to get rid of the array reverse above.

grassroots-backend/src/mikro-orm.config.ts
tdresser: That will require exporting a promise. This isn't well documented, but sounds doable: https://github.com/mikro-orm/mikro-orm/discussions/4185

If this proves to be too annoying, don't worry about it.

tdresser (commented):
This is great, thanks! Just suggested some minor polish.

However, looks like some tests are failing, and I suspect it's a real issue.

Can you reproduce it locally? You might need to look at how the test / ci env files relate, and construct a similar setup locally.

grassroots-backend/src/mikro-orm.config.ts
MuhammadElsoukkary: Looks like I got some issues with the codebase expecting sync when trying to implement it as async  image
image

tdresser: We should be able to address those failures by (sorry for the bad variable names, I'm on mobile...):

import mikroormConfigPromise from ...


...


const mikroormconfig = await mikroormConfigPromise.

And then access values off that.

PR #159, opened 2025-07-13
Add eslint rules for dto and entity definition and usage

tdresser:
Enforces class name case, and that DTO / Entity must be at the end of the class name.
No constructors allowed in DTOs or entities.
No variables declared with entity types outside of services.

### DIFF
--- a/eslint_rules/src/rules/dto-and-entity-style.test.ts
+++ b/eslint_rules/src/rules/dto-and-entity-style.test.ts
@@ -1,27 +1,7 @@
-import path from "node:path";
-import tseslint from "typescript-eslint";
-import { RuleTester } from "@typescript-eslint/rule-tester";
-import * as vitest from "vitest";
-
 import { rule } from "./dto-and-entity-style.js";
+import { createRuleTester } from "../utils.js";
 
-RuleTester.afterAll = vitest.afterAll;
-RuleTester.it = vitest.it;
-RuleTester.itOnly = vitest.it.only;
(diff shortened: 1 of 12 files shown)


eslint_rules/src/rules/entity-use.ts
```diff
@@ -0,0 +1,61 @@
...
+      messageId: "noEntityAccessOutsideServices",
+      node: node,
+    });
+  }
+}
+
+export const rule = createRule({
+  create(context: Context): RuleListener {
+    return {
+      VariableDeclaration(node: TSESTree.VariableDeclaration): undefined {
```
tdresser, line 27: TSTypeReference would be better to use here.

eslint_rules/src/rules/dto-and-entity-style.test.ts
```diff
@@ -44,5 +24,47 @@ ruleTester.run("definite-or-optional", rule, {
...
       ],
     },
+    {
+      code: `class FooDto { a!: number}`,
+      errors: [
+        {
+          column: 1,
+          endColumn: 27,
+          line: 1,
+          endLine: 1,
```
IanEdington, line 34: Had you mentioned these were hard to maintain?

IanEdington (approved):

eslint_rules/src/rules/dto-and-entity-style.test.ts
tdresser: I'll get rid of this in a follow up.

PR #160, opened 2025-07-13
pre-commit hook sometimes doesn't run reliably

markwong167:
the ./run makes it slower and causes issues

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-./run npx lint-staged
+npx lint-staged


markwong167:
> Would you be up for adding a parameter letting us skip the ensure_container_running() step, and using that here?
> 
> I think that should eliminate the slowness and flake.

I am skipping the whole "run" script, which includes ensure_container_running.
Was there something in "run" that was useful or necessary?
If so and if it's small enough, I would rather just call that in "pre-commit" rather than using "run", as using "run" feels like it's over-complicating the pre-commit.

tdresser:
@MuhammadElsoukkary might have thoughts!

MuhammadElsoukkary:
As for slowness I think that was an issue me and Ian were having when testing the script, I was also getting the same this is a not git repo issue however I do not know what is the culprit behind it.

PR #161, opened 2025-07-14
Adding pnpm store to gitignore to prevent craziness

markwong167:
There will be a .pnpm-store that will store thousands of json pointer files to speed things up.
having a gitignore line seems to be good and standard practice


tdresser:
I'd tend to just merge this in the same PR as migrating to pnpm. Happy to approve if this meaningfully simplifies things though.

Let me know!

PR #164, opened 2025-07-15
A somewhat aggressive approach to type branding

tdresser:
Typescript has "structural" typing, which means that two types that are the same shape will be automatically converted.

Consider this example:
```
class Foo {}

class ADTO {
    a!: number
    fooId?: number
}

class AEntity {
    a!:number
    foo?: Foo
}

let aDTO:ADTO = {a:3};
let aEntity: AEntity = aDTO;
```

Typescript allows this, despite it being clearly wrong.
To avoid this, we introduce a compile time only "brand", used
only to convince typescript that these types shouldn't be castable to one another.

We also use this base type to provide common methods on DTOs and Entities, such as all Entities needing a "toDTO" method.

In addition to the regular benefits of stronger typing, this also eliminates a source of "I just wanted to add a tiny helper method, and now I need to update 40 files".
Many of our DTOs and Entities are likely to slowly deviate to the point where we won't get implicit casting, even without the brand. When that happens, you'll then need to add an explicit cast everywhere it's required, which is more of a pain to do after the fact than when you initially start using the entity and DTO.

The brand ensures we're already handling this, so you do a little bit of work up front, but don't suddenly need to add a bunch of explicit casts just because you added a helper method to a DTO.

Also, we'll need explicit casts sometimes, and it's nicer to be 100% consistent, as opposed to only explicitly casting when the set of properties diverges too far.

### DIFF
--- a/grassroots-backend/src/grassroots-shared/util/CreateDTOBase.ts
+++ b/grassroots-backend/src/grassroots-shared/util/CreateDTOBase.ts
@@ -30,9 +30,11 @@ import { PropsOf } from "./PropsOf";
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export function createDTOBase<TBrand extends string>(brand: TBrand) {
   abstract class Branded {
-    readonly __brand!: `${TBrand}DTO`;
+    // We need different names for our branding types
+    // between entities or DTOs, or we get some type collicions.
+    readonly __DTOBrand!: `${TBrand}DTO`;
     // Used for CASL to identify object types.
-    readonly __caslSubjectType__ = brand;
+    readonly __caslSubjectType: string = brand;
 
     static from<T extends Branded>(
       // The this parameter must be named "this", and is magically populated with the class constructor.
--- a/grassroots-backend/src/grassroots-shared/util/PropsOf.ts
+++ b/grassroots-backend/src/grassroots-shared/util/PropsOf.ts
@@ -2,7 +2,7 @@
 // Recursively does this with all included objects.
 
 // Exclude these keys which are used for internal bookkeeping.
-type ExcludedKeys = "__brand" | "__caslSubjectType__";
+type ExcludedKeys = "__DTOBrand" | "__entityBrand" | "__caslSubjectType";
 
 export type PropsOf<T> = {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
--- a/grassroots-backend/src/testing/PropsOf.spec.ts
+++ b/grassroots-backend/src/testing/PropsOf.spec.ts
@@ -1,17 +1,15 @@
 import { describe, expect, it } from "vitest";
 import { PropsOf } from "../grassroots-shared/util/PropsOf";
 
-const __brand: unique symbol = Symbol();
-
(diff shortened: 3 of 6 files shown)


grassroots-backend/src/grassroots-shared/util/PropsOf.ts
```diff
@@ -1,5 +1,28 @@
 // Roughly, given a class, gives you an interface with its attributes.
+// Recursively does this with all included objects.
+
+// Exclude these keys which are used for internal bookkeeping.
+type ExcludedKeys = "__brand" | "__caslSubjectType__";
+
 export type PropsOf<T> = {
```
IanEdington, line 7: Thanks for the comments. This is very well laid out

grassroots-backend/src/testing/PropsOf.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+class Branded<TBrand> {
+  readonly [__brand]!: TBrand;
+  next?: Branded<TBrand>;
+}
+
+describe("propsOf", () => {
+  it("should exclude brands", () => {
+    const x: PropsOf<Branded<"foo">> = { next: undefined };
+    expect(x.next).toStrictEqual(undefined);
```
IanEdington, line 14: I'm not sure I understand this test. I would have expected `expect(x).toStrictEqual({next:undefined})`. Maybe I don't understand what's begins tested here.

grassroots-frontend/src/routes/Search.tsx
```diff
@@ -30,15 +30,19 @@ function Search(): JSX.Element {
 
   const [rowsToSkip, setRowsToSkip] = useState<number>(0);
 
-  const searchParams: PaginatedContactSearchRequestDTO = {
-    contact: plainToInstance(ContactSearchRequestDTO, form.watch()),
+  const searchParams = PaginatedContactSearchRequestDTO.from({
+    contact: form.watch(),
```
IanEdington, line 34: I REALLY like how this approach makes it clear when it needs to be used compared to "plain to instance" which lacks context of why it's need, so make it hard to know when it's needed.

grassroots-backend/src/grassroots-shared/util/CreateDTOBase.ts
```diff
@@ -0,0 +1,46 @@
...
+export function createDTOBase<TBrand extends string>(brand: TBrand) {
+  abstract class Branded {
+    readonly __brand!: `${TBrand}DTO`;
+    // Used for CASL to identify object types.
+    readonly __caslSubjectType__ = brand;
+
+    static from<T extends Branded>(
+      // The this parameter must be named "this", and is magically populated with the class constructor.
+      this: new () => T,
+      props: PropsOf<T>,
```
IanEdington, line 40: TIL about `this`

IanEdington (approved):
😍 wow this is a work of beauty!

grassroots-backend/src/testing/PropsOf.spec.ts
tdresser: You're right! This is busted!

Fixed.

PR #170, opened 2025-07-17
Migrate from npm to pnpm

markwong167:
this migrates from npm to pnpm and allows use on the dev container.
Fully creating the workspace comes later.


tdresser:
Happy to take a look at the current error when I get the chance. 

I do think we'll want to figure out shimming before landing. I'll probably land code that depends on shimming within the next few days

PR #171, opened 2025-07-20
BDD tests (WIP)

tprost:
You can read about acceptance testing here

[https://cucumber.io/](https://cucumber.io/)

In a nutshell, tests/scenarios are written in a plain English, which double as documentation on how the application should behave, and we take those descriptions and turn them into code that can test our app e2e.

Some things I'm thinking about:

1. We probably want a test Google account that we can use in these tests so we can test the login flow (and the rest of the app for that matter)
2. I need to read up on this [https://playwright.dev/docs/auth#introduction](https://playwright.dev/docs/auth#introduction)
3. I need to wipe and reset the DB every time these tests run, and if we want to use that separate test_db, we need to run another copy of the backend locally specifically for testing
4. If we run these tests in Docker, you won't be able to see the browser do stuff, which is kind of nice to be able to do (you can take screenshots though)


tprost:
The cute thing about Cucumber is Ian can write up exactly how the app is supposed to behave in plain English, which is great documentation for beginners and newcomers, and he's also helping us write tests :)

tprost:
Using a different branch

PR #182, opened 2025-08-02
Give each contact an organization

tdresser:
This PR doesn't add any UX, so for the frontend, we temporarily just say that passing an id of -1 will magically create an organization when creating a contact.

This is ugly, but temporary...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@ node_modules
 grassroots-backend/dist
 grassroots-backend/src/metadata.ts
 grassroots-frontend/dist
-grassroots-frontend/src/**/*.js
 grassroots-frontend/vite.config.js
 .env.development.local
 .env.test.local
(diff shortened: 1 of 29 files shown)


.gitignore
```diff
@@ -2,7 +2,6 @@ node_modules
 grassroots-backend/dist
 grassroots-backend/src/metadata.ts
 grassroots-frontend/dist
-grassroots-frontend/src/**/*.js
```
tdresser, line 5: These files are annoying, and this makes it easy to just git clean them away.

IanEdington (approved):
looks good. LGTM

.gitignore
tdresser: They currently should only be put in the dist folder. I just accidentally do a raw tsc forgetting --noEmit rather regularly...

PR #189, opened 2025-08-10
CASL part 1

tdresser:
This implements CASL authorization, but doesn't use it, or integrate it at the db level.
The framework for using it with mikroORM is landed, but only used in a test showing off roughly how it works.

Followup patches will integrate this into our repository and baseentitiy classes, and ensure this is used by our services.

### DIFF
--- a/grassroots-backend/src/grassroots-shared/Permission.ts
+++ b/grassroots-backend/src/grassroots-shared/Permission.ts
@@ -1,0 +1,172 @@
+import {
+  AbilityBuilder,
+  createMongoAbility,
+  MongoAbility,
+  MongoQuery,
+  subject,
+} from "@casl/ability";
+import { UserEntity } from "../users/User.entity";
+import { ContactEntity } from "../contacts/entities/Contact.entity";
+import { CommonProps } from "./util/TypeUtils";
+import { AbilityQuery, rulesToQuery } from "@casl/ability/extra";
+import { FilterQuery } from "@mikro-orm/core";
+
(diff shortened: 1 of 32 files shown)


grassroots-backend/src/grassroots-shared/Permission.ts
```diff
@@ -0,0 +1,81 @@
...
+export function can(
+  ability: AppAbility,
+  action: CASLAction,
+  type: keyof CASLSubjects,
+  object: CASLSubjectUnion,
+): boolean {
+  return ability.can(action, subject(type, object));
+}
+
+export function permissionsToCaslAbilities(
```
tdresser, line 50: This should probably be refactored out to a separate file which only declares the rules.

grassroots-frontend/src/routes/Users.tsx
```diff
@@ -12,10 +12,11 @@ export const Route = createFileRoute("/Users")({
...
   const { data: users } = useQuery<UsersDTO>({
     queryKey: ["users"],
+    initialData: UsersDTO.from({ users: [] }),
     queryFn: async () => {
       const result = await grassrootsAPI.GET("/users", {});
       return UsersDTO.fromFetchOrThrow(result);
     },
   });
-  return <div>{(users?.users ?? []).map((x) => UserRow({ user: x }))}</div>;
+  return <div>{users.users.map((x) => UserRow({ user: x }))}</div>;
```
tdresser, line 21: TODO

PR #191, opened 2025-08-29
Admin page init

markwong167:
Created admin page with basic table and sample data


tdresser (approved):
LGTM, with one annoying bit of feedback. 
I'm not sure "admin" is the right name for this page. 

Admin could just as easily mean user management, for example. 

What do you think about "organizations" as the name? Or something like that?

markwong167:
I'm imagining organizations to be something separate, like something that could be accessed by many roles.
This page has the purpose of organization management, but that would be an insanely long name.
I think Organizers would be the best name here, what do you think?
The title would be Organizer Dashboard

PR #193, opened 2025-09-01
Org dashboard

markwong167:
created the start of an org dashboard for organizers to manage it

### DIFF
--- a/grassroots-frontend/src/pages/me/MyTable.tsx
+++ b/grassroots-frontend/src/pages/me/MyTable.tsx
@@ -8,16 +8,15 @@ interface OrganizersTableProps {
 
 export function MyTable({ tableData }: OrganizersTableProps): JSX.Element {
   const rows = tableData.map((riding) => (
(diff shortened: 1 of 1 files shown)


tdresser (commented):
I think what you're calling an organizer is what we're calling a "User" currently. Does that sound right? 
 
Which name seems better to you?

Could you use an array of UserDTO objects for your mock data?

markwong167:
Here's how I'm thinking about it.
Not all users are organizers, but all organizers are users.
But all users are members and vice versa here.
Perhaps it might be better to do it as
Route: "User"
User Page Title: "Your Dashboard"
User Table: "Your Organizations"
Separate Route: "Organizations/$OrgAbbr"
Org Page title: [Name of Org]
Org Table: "Members" but under the hood, it's Users
Then if the role for that organization is "Manager", they have Create,Read, Update, and Deactivate capabilities in both the Org Table and the User Table.

markwong167:
Oh right, I forgot to have the sample data follow the DTO's.

PR #203, opened 2025-09-09
Everything is a pnpm package

tdresser:
Once you try to make one thing a package, you kind need to make everything a package.

This is unpleasantly large, but I don't think it's worth splitting it up, given our current focus on switching from infra to feature work.

Seem reviewable?

Start with README_PACKAGES.md.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
         run: pnpm run -r build
 
       - name: Migrate DB
-        run: cd grassroots-backend && pnpm run db-up
+        run: cd grassroots-backend && pnpm exec mikro-orm-esm migration:up
 
       - name: Setup tbls
         uses: k1low/setup-tbls@v1
@@ -108,7 +108,7 @@ jobs:
 
       - name: Check mikro-orm migration
         run: |
-          if ! cd grassroots-backend && pnpm run db-check 2>&1 | grep -q "No changes required, schema is up-to-date"; then
+          if ! cd grassroots-backend && pnpm run pnpm exec mikro-orm-esm migration:check 2>&1 | grep -q "No changes required, schema is up-to-date"; then
             exit 1
           fi
 


generate-metadata/README.md
```diff
@@ -0,0 +1,4 @@
+This package's `build` command generates `/app/grassroots-shared/src/metadata.ts`.
```
tdresser, line 1: Er, this isn't correct yet, but it will be next patch.

.syncpackrc.json
```diff
@@ -0,0 +1,22 @@
...
+        "**"
+      ],
+      "dependencies": [
+        "$LOCAL"
+      ],
+      "dependencyTypes": [
+        "dev",
+        "prod"
+      ],
+      "pinVersion": "workspace:*"
```
tdresser, line 19: Without this, syncpack breaks workspace dependencies.

.github/workflows/ci.yml
```diff
@@ -109,19 +95,22 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: "Services: Start Test Services"
+        uses: hoverkraft-tech/compose-action@v2.2.0
+        with:
+          compose-file: "./docker/compose-ci.yaml"
+
+      - name: Build everything
```
tdresser, line 103: We shouldn't need to build everything here, but it's the easiest way to make sure all the dependencies of the tests are build.

turborepo will clean this up.

generate-metadata/src/GenerateMetadata.ts
```diff
@@ -0,0 +1,32 @@
...
+const METADATA_TMP_FILENAME = "metadata.withbadpaths.ts";
+const METADATA_TMP_PATH = METADATA_TMP_DIR + METADATA_TMP_FILENAME;
+
+const generator = new PluginMetadataGenerator();
+generator.generate({
+  visitors: [
+    new ReadonlyVisitor({
+      introspectComments: true,
+      pathToSource: "./src",
+      esmCompatible: true,
```
tdresser, line 17: esmCompatible is the special sauce to not need my nonsense regexes!

generate-metadata/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "generate-metadata",
+  "version": "1.0.0",
+  "description": "",
+  "main": "GenerateMetadata.js",
+  "type": "module",
+  "scripts": {
+    "build": "rm -f /app/grassroots-backend/src/grassroots-shared/metadata.ts && swc src --strip-leading-paths --out-dir dist && node dist/GenerateMetadata.js"
```
tdresser, line 8: Calling this build is a bit weird. Again, we could fix this, but it probably makes more sense to fix it once we're using turborepo. 

I didn't want to introduce turborepo in this patch, as it's already huge.

grassroots-backend/package.json
```diff
@@ -5,36 +5,48 @@
...
-    "build": "nest build",
+    "build": "nest build && tsc -p tsconfig.buildwithtypes.json",
     "start": "nest start --watch",
-    "start:prod": "node dist/Main",
+    "start:prod": "node dist-build/Main",
     "gen-files-only": "nest start -- --gen-files-only",
-    "debug": "pnpm run build && node --inspect dist/Main",
+    "debug": "pnpm run build && node --inspect dist-build/Main",
+    "test:tstyche": "pnpm exec tstyche",
+    "db-create": "tsc -p tsconfig.mikroorm.json && pnpm exec mikro-orm migration:create --config dist/mikro-orm.config.js",
```
tdresser, line 22: The fact that the command needs to come before the "--config ..." is... terrible?

Otherwise I'd just create a single db command, and you could `pnpm run db -- migration:create`

We could create a wrapper script, but I think this is fine for now.

grassroots-frontend/src/grassroots-shared
tdresser, line 1: GONE FOREVER.

grassroots-backend/package.json
tdresser: Fixed this!

PR #212, opened 2025-09-19
Add Module For Phone Canvass

tdresser:
Just a skeleton of the phone canvass module.
Mostly just a `pnpm exec nest generate resource phone-canvas`.

Also updated grassroots-e2e-tests to depend on grassroots-backend via source, not a built package.
Might want to do that more broadly...


grassroots-shared/src/dtos/PhoneCanvass/CallStatus.dto.ts
```diff
@@ -0,0 +1,22 @@
+import { IsEnum } from "class-validator";
```
tdresser, line 1: This does the exact same thing as the Permission enum.

PR #214, opened 2025-09-22
Hide current routes, add route for phone canvass

tdresser:
I'm going to need to drop some test code into the frontend for twilio at some point, and don't want to get conflicts with Mark, so figured I'd drop a placeholder route in for now.

### DIFF
--- a/grassroots-frontend/src/routes/__root.tsx
+++ b/grassroots-frontend/src/routes/__root.tsx
@@ -31,16 +31,9 @@ export const Route = createRootRouteWithContext<RouterContext>()({
             <RoutedLink to="/">Home</RoutedLink>
           </AppShell.Section>
           <AppShell.Section>
-            <RoutedLink to="/CreateContact">Create Contact</RoutedLink>
-          </AppShell.Section>
-          <AppShell.Section>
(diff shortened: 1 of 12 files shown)


grassroots-frontend/src/routes/__root.tsx
```diff
@@ -31,16 +31,9 @@ export const Route = createRootRouteWithContext<RouterContext>()({
...
           </AppShell.Section>
           <AppShell.Section>
-            <RoutedLink to="/CreateContact">Create Contact</RoutedLink>
-          </AppShell.Section>
-          <AppShell.Section>
-            <RoutedLink to="/Search">Search Contacts</RoutedLink>
-          </AppShell.Section>
-          <AppShell.Section>
-            <RoutedLink to="/Users">Users</RoutedLink>
-          </AppShell.Section>
```
markwong167, line 41: Let's keep /Me
That's for Admins of some sort, if I recall correctly

tdresser: Happy to leave it. 
Any thoughts on what it would do in the MVP?

PR #216, opened 2025-09-23
Fix turbo build, improve performance some. 

tdresser:


grassroots-backend/src/util/LastDependencyUpdateTime.ts
```diff
@@ -0,0 +1,2 @@
+// This file is updated when grassroots-shared is done building to trigger a rebuild in watch mode.
+export const LAST_DEPENDENCY_UPDATE_TIME = 209989.771846;
```
IanEdington, line 2: Can this be git ignored? I suspect if it's not it will change in every PR going forward

IanEdington (approved):

PR #224, opened 2025-09-28
Twilio connecting phone and browser

tdresser:
This is just the proof of concept, it's not really properly wired in yet, but I think it's probably worth reviewing at this stage.
Removing the placeholders requires a fair bit more work, and I don't want to send out a ginormous patch.

I have my phone and browser communicating (though I had to whitelist my number in twilio)!


grassroots-backend/src/phone-canvass/PhoneCanvass.service.ts
```diff
@@ -1,18 +1,24 @@
...
+import { TwilioService } from "./Twilio.service.js";
+import { VoidDTO } from "grassroots-shared/dtos/Void.dto";
 
 @Injectable()
 export class PhoneCanvassService {
   repo: EntityRepository<PhoneCanvassEntity>;
-  constructor(private readonly entityManager: EntityManager) {
+  constructor(
+    private readonly entityManager: EntityManager,
+    private twilioService: TwilioService,
```
tdresser, line 20: General model I'm trying to follow here is that the PhoneCanvasService relies on the TwilioService for anything that directly hits Twilio, so at some point we can mock out the TwilioService completely.

grassroots-backend-e2etests/src/PhoneCanvass.e2e-spec.ts
```diff
@@ -62,4 +64,22 @@ describe("PhoneCanvass (e2e)", () => {
...
+  it("should provide valid twiml from the webhook", async () => {
+    const f = getFixture();
+
+    const params = new URLSearchParams();
+    params.append("conference", "test");
+
+    const result = await f.grassrootsAPIRaw("/phone-canvass/twilio-voice", {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
```
tdresser, line 77: This is the way twilio sends webhook information.

PR #234, opened 2025-10-03
DTO::from magically converts entities to DTOs

tdresser:
I keep accidentally losing fields when I forget to do this conversion, so this automates it.

This is just a runtime thing, the type signature is unchanged, which is a bit weird, but the properties of DTOs and entities are sufficiently similar that I think this is pretty reasonable.

### DIFF
--- a/grassroots-shared/tsconfig.json
+++ b/grassroots-shared/tsconfig.json
@@ -3,5 +3,5 @@
   "compilerOptions": {
     "outDir": "./dist"
   },
-  "include": ["src", "../grassroots-backend/src/testing/CreateDTOBase.spec.ts"]
+  "include": ["src"]
 }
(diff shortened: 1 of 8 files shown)


grassroots-shared/tsconfig.json
```diff
@@ -3,5 +3,5 @@
   "compilerOptions": {
     "outDir": "./dist"
   },
-  "include": ["src"]
+  "include": ["src", "../grassroots-backend/src/testing/CreateDTOBase.spec.ts"]
```
IanEdington, line 6: are there best practices regarding crossing the package boundary and when it's OK?

tdresser: I have no idea why I was crossing the package boundary here. This seems completely unnecessary...

Removed it, and I'll see if anything explodes.

In general, I think the best practice is to never cross the package boundary, but in practice it sometimes makes things significantly simpler. (But not here, as far as I can tell)

PR #235, opened 2025-10-03
Add Call Party page

tdresser:


grassroots-frontend/src/pages/organizations/OrganizationPage.tsx
```diff
@@ -11,7 +10,7 @@ import { PaginatedContacts } from "../../components/PaginatedContacts.js";
 const ROWS_PER_PAGE = 10;
 
 export function OrganizationDashboard(): JSX.Element {
-  const { organizationId } = useParams({ from: OrgRoute.id });
+  const { organizationId } = OrgRoute.useParams();
```
tdresser, line 13: I wasn't getting strong typing of params without this approach.

markwong167: Okay, I don't have a strong opinion on this

grassroots-frontend/src/routes/PhoneCanvass/$phoneCanvassId.tsx
```diff
@@ -0,0 +1,22 @@
+import { createFileRoute } from "@tanstack/react-router";
+import { JSX } from "react";
+import { StartCall } from "../../components/phonecanvass/StartCall.js";
```
markwong167, line 3: it seems weird that we keep saying .js at the end of .tsx files
It should work without.
I'm also more of a fan of naming conventions where we can read the folder name easily, but it's not a big thing

grassroots-frontend/src/routes/PhoneCanvass/Manage.$phoneCanvassId.tsx
```diff
@@ -33,6 +34,12 @@ function ManagePhoneCanvass(): JSX.Element {
   return (
     <>
       <h1> Manage your phone canvass </h1>
+      <Link
+        to="/PhoneCanvass/$phoneCanvassId"
+        params={{ phoneCanvassId: phoneCanvassId }}
+      >
+        <Button>Participate</Button>
+      </Link>
```
markwong167, line 42: My understanding is that a button inside of a link is an antipattern.
The two things that I do are:
1. Create a button-like link component
2. Just use the button's onclick to do something, then navigate
Depends on the situation

grassroots-frontend/src/routes/PhoneCanvass/$phoneCanvassId.tsx
tdresser: Re: .js at the end of .tsx files, I _think_ this is the least of all evils, but I also don't like it.
To get the monorepo working, I had to migrate to NodeNext on the backend, which requires the .js extension for .ts files (that are being included via relative paths). For consistency, I think it makes sense to do this throughout the project, so I added an eslint rule enforcing it. This makes things like moving code between backend and frontend a bit less painful.

We could special case .tsx files to not require the suffix, but that seems a bit weird to me.

"naming conventions where we can read the folder name easily"
Any concrete suggestions? Not sure why I didn't go with /components/PhoneCanvass/StartCall.js, is that what you mean, or something else?

markwong167: Yeah, PhoneCanvass is the best, I think.
My guess is that naming folders exactly like files might have felt taboo for you. Especially when you have a PhoneCanvass somewhere else.

markwong167 (approved):
LGTM

PR #238, opened 2025-10-05
Add pre-call screen

tdresser:


grassroots-backend/src/auth/Auth.controller.ts
```diff
@@ -114,6 +115,9 @@ export class AuthController {
   async getActiveOrg(
     @Session() session: SessionData,
   ): Promise<OrganizationDTO> {
+    if (session.activeOrganizationId == undefined) {
```
tdresser, line 118: Previously this was considered always defined, which was obviously wrong.

grassroots-backend/src/contacts/entities/Contact.entity.ts
```diff
@@ -21,11 +15,9 @@ export class ContactEntity extends createEntityBase<"Contact", ContactDTO>(
   id!: number;
 
   @Property({ nullable: true })
-  @Unique()
   gvote_id?: string;
```
tdresser, line 18: As discussed, these aren't unique!

grassroots-backend/src/phone-canvass/PhoneCanvass.service.ts
```diff
@@ -120,4 +125,61 @@ export class PhoneCanvassService {
...
+    const partitionedContacts = partition(contacts, (contact) => {
+      if (contact.callStatus === "NOT_STARTED") {
+        return "NOT_STARTED";
+      } else if (contact.callStatus === "STARTED") {
+        return "STARTED";
+      }
+      return "COMPLETE";
+    });
+
+    const activeCalls: ActiveCall[] = (
```
tdresser, line 141: These bits are still untested at this point. My next step is probably to start writing some tests which simulate actually making calls, which will cover this.

grassroots-backend/src/phone-canvass/Twilio.service.ts
```diff
@@ -110,28 +89,26 @@ export class TwilioService {
...
-    const client = this.getClient();
+    const client = this.#getClient();
 
-    let doc: undefined | DocumentInstance;
     try {
       // This fails if the document doesn't already exist. Checking if it exists first introduces
       // an extra unnecessary round trip and some complexity.
       // Instead we just try to update it, and if that fails, create it instead.
-      doc = await client.sync.v1
+      await client.sync.v1
```
tdresser, line 104: We weren't using these return values, and they made writing mocks much more annoying.

grassroots-shared/src/decorators/Trim.decorator.ts
```diff
@@ -0,0 +1,19 @@
+import { Transform } from "class-transformer";
```
tdresser, line 1: I want participant display names to be unique without worrying about whitespace.

PR #258, opened 2025-10-20
Initial scheduler implementation

tdresser:
This isn't integrated at all, and is missing some meaningful bits (like writing state to contact entities), but I think there's enough here to be worth reviewing and landing.

It's already pretty large, but I think it's tricky to break this up while still making the value of each piece clear.

### DIFF
--- a/grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassCall.ts
+++ b/grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassCall.ts
@@ -4,7 +4,7 @@ import {
   CallStatus,
 } from "grassroots-shared/dtos/PhoneCanvass/CallStatus.dto";
 import { PhoneCanvassContactEntity } from "../entities/PhoneCanvassContact.entity.js";
-import { PhoneCanvassSchedulerImpl } from "./PhoneCanvassScheduler.js";
+import { PhoneCanvassScheduler } from "./PhoneCanvassScheduler.js";
 
 export type Call =
   | NotStartedCall
@@ -16,7 +16,7 @@ export type Call =
 
 interface CommonCallState {
   id: number;
-  scheduler: PhoneCanvassSchedulerImpl;
+  scheduler: PhoneCanvassScheduler;
   contact: PhoneCanvassContactEntity;
   // Timestamps provided by Twilio when available.
   // Otherwise, we use Date.now().
--- a/grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassMetricsTracker.ts
+++ b/grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassMetricsTracker.ts
@@ -1,11 +1,11 @@
 import { BehaviorSubject, combineLatest, map, Observable, Subject } from "rxjs";
 import { Call, CompletedCall } from "./PhoneCanvassCall.js";
-import { PhoneCanvassSchedulerImpl } from "./PhoneCanvassScheduler.js";
+import { PhoneCanvassScheduler } from "./PhoneCanvassScheduler.js";
 
 export class PhoneCanvassMetricsTracker {
   #endingCallsObservable = new Subject<CompletedCall>();
   #callerCountObservable = new BehaviorSubject<number>(0);
-  #committedCallerCountObservable = new BehaviorSubject<number>(0);
+  #committedCallsCountObservable = new BehaviorSubject<number>(0);
 
   readonly #idleCallerCountObservable: Observable<number>;
 
@@ -18,7 +18,7 @@ export class PhoneCanvassMetricsTracker {
   }
 
   get committedCallerCountObservable(): Observable<number> {
-    return this.#committedCallerCountObservable;
+    return this.#committedCallsCountObservable;
   }
 
   get idleCallerCountObservable(): Observable<number> {
@@ -28,11 +28,11 @@ export class PhoneCanvassMetricsTracker {
   constructor() {
     this.#idleCallerCountObservable = combineLatest([
       this.#callerCountObservable,
-      this.#committedCallerCountObservable,
+      this.#committedCallsCountObservable,
     ]).pipe(
       map(
-        ([callerCount, committedCallerCount]) =>
-          callerCount - committedCallerCount,
+        ([callerCount, committedCallsCount]) =>
+          callerCount - committedCallsCount,
       ),
     );
 
@@ -46,11 +46,9 @@ export class PhoneCanvassMetricsTracker {
   }
 
   onCallsByStatusUpdate(
-    callsByStatus: InstanceType<
-      typeof PhoneCanvassSchedulerImpl
-    >["callsByStatus"],
+    callsByStatus: PhoneCanvassScheduler["callsByStatus"],
   ): void {
-    this.#committedCallerCountObservable.next(
+    this.#committedCallsCountObservable.next(
       callsByStatus.NOT_STARTED.size +
         callsByStatus.INITIATED.size +
         callsByStatus.QUEUED.size +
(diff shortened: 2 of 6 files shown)


grassroots-shared/src/dtos/PhoneCanvass/CallStatus.dto.ts
```diff
@@ -20,3 +25,20 @@ export function CallStatusDecorator(): PropertyDecorator {
     ApiProperty({ enum: CallStatusEnum }),
   );
 }
+
+enum CallResultEnum {
+  CANCELED = "CANCELED",
+  COMPLETED = "COMPLETED",
```
tdresser: I don't love "COMPLETED" being both a result and a status, but it aligns with what twilio does.

grassroots-shared/src/util/DeepLog.ts
```diff
@@ -0,0 +1,13 @@
+const replacer: (this: unknown, key: string, value: unknown) => unknown = (
```
tdresser: I keep trying to Log maps and having it break.
I think landing this helper is probably reasonable, even though I'm not actually using it in the code landed in this patch.

grassroots-shared/src/dtos/PhoneCanvass/CallStatus.dto.ts
IanEdington: what do COMPLETED and FAILED mean in this case? Maybe a link to the twilio docs would be helpful on this enum?

grassroots-shared/src/dtos/PhoneCanvass/CallStatus.dto.ts
```diff
@@ -20,3 +25,20 @@ export function CallStatusDecorator(): PropertyDecorator {
...
+  BUSY = "BUSY",
+  NO_ANSWER = "NO_ANSWER",
+  FAILED = "FAILED",
+}
+
+export type CallResult = keyof typeof CallResultEnum;
+
+export function CallResultDecorator(): PropertyDecorator {
```
IanEdington: I'm not seeing where this is used. Was the idea to use this in the Entity?

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassScheduler.ts
```diff
@@ -0,0 +1,153 @@
+/* eslint-disable grassroots/entity-use */
```
IanEdington: I see this turned off in a number of files. Is there something wrong with the rule? Should we disable it globally?
Usually I prefer having linter exceptions at the line level so it's easier to see and remove them in the future.

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassCall.ts
```diff
@@ -0,0 +1,227 @@
...
+    return call;
+  }
+
+  get id(): number {
+    return this.state.id;
+  }
+}
+
+export class NotStartedCall extends AbstractCall<"NOT_STARTED"> {
+  static #currentId = 0;
```
IanEdington: Is the idea that every call would have a unique ID? Is there a reason this should be different from the PhoneCanvassContactEntity id?

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassScheduler.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    return oldestAvailableCaller.id;
+  }
+
+  addCaller(id: number): void {
+    this.#callers.set(id, {
+      id,
+      availabilityStartTime: this.getCurrentTime(),
+    });
+    this.metricsTracker.onCallerCountUpdate(this.#callers.size);
```
IanEdington: Rather than have metrics code sprinkled through the call code, would it be possible to have metrics observe the call observables? I know that probably won't always work

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassScheduler.ts
```diff
@@ -0,0 +1,153 @@
...
+    this.#strategy = new NoOvercallingStrategy(this.metricsTracker);
+  }
+
+  async start(): Promise<void> {
+    if (this.#running) {
+      throw new Error("Call to startScheduler when scheduler already running.");
+    }
+    this.#running = true;
+    while (this.#running) {
+      await this.#strategy.waitForNextCall();
```
IanEdington: because of the await, I believe `stop` will allow one last call to go through. Maybe not a big deal, but we could check after the await `if (!this.#running) break;`

grassroots-shared/src/dtos/PhoneCanvass/CallStatus.dto.ts
tdresser: This will be required to pass a CallResult to the frontend. Maybe we won't need this though?

Will call party participants want to know if a call was removed from the queue because it was busy etc?

That does sound probably post MVP, in which case I can remove this.

grassroots-shared/src/dtos/PhoneCanvass/CallStatus.dto.ts
tdresser: Yeah, my original implementation actually did lump them together. 

I think they are actually conceptually pretty distinct though. The state machine doesn't care why a call ended, it just needs to know that it did. 

We only need to know the result for logging / statistics purposes. 

I kinda like splitting them out for that reason.

Also, with the way I've implemented the state transitions, it would mean either a bit more complexity (one class with multiple possible states) or a bunch more boilerplate. 

The conceptual distinction is a much stronger reason though. :P

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassMetricsTracker.ts
```diff
@@ -0,0 +1,65 @@
...
+    });
+  }
+
+  onEndingCall(call: CompletedCall): void {
+    this.#endingCallsObservable.next(call);
+  }
+
+  onCallsByStatusUpdate(
+    callsByStatus: InstanceType<
+      typeof PhoneCanvassSchedulerImpl
```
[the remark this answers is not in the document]
tdresser: The interface doesn't include callsByStatus, as it's generally an implementation detail. I'd prefer clients couldn't reach in and grab it. 

I've exported the type separately to avoid needing to depend on the implementation.

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassCall.ts
```diff
@@ -0,0 +1,227 @@
...
+  | NotStartedCall
+  | QueuedCall
+  | InitiatedCall
+  | RingingCall
+  | InProgressCall
+  | CompletedCall;
+
+interface CommonCallState {
+  id: number;
+  scheduler: PhoneCanvassSchedulerImpl;
```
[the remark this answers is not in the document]
tdresser: It previously couldn't because callsByStatus wasn't exposed. I've exposed it.

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassMetricsTracker.ts
```diff
@@ -0,0 +1,65 @@
+import { BehaviorSubject, combineLatest, map, Observable, Subject } from "rxjs";
+import { Call, CompletedCall } from "./PhoneCanvassCall.js";
+import { PhoneCanvassSchedulerImpl } from "./PhoneCanvassScheduler.js";
+
+export class PhoneCanvassMetricsTracker {
+  #endingCallsObservable = new Subject<CompletedCall>();
+  #callerCountObservable = new BehaviorSubject<number>(0);
+  #committedCallerCountObservable = new BehaviorSubject<number>(0);
```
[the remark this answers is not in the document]
tdresser: Oh, yeah, that's way better, thanks!

grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassScheduler.ts
tdresser: Great point. eslint doesn't realize that the value could change during that await, so I needed to
```// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition```
which is a bit scary.

PR #259, opened 2025-10-26
Rename participant to caller, give callers an id

tdresser:


.devcontainer.json
```diff
@@ -7,8 +7,7 @@
         "bierner.markdown-mermaid",
         "ms-azuretools.vscode-docker",
         "github.vscode-github-actions",
-        "editorconfig.editorconfig",
-        "Vercel.turbo-vsc"
```
tdresser: The turbo extension doesn't actually work with docker AFAICT.

grassroots-backend/src/phone-canvass/PhoneCanvassGlobalState.service.ts
```diff
@@ -3,55 +3,51 @@ import {
...
     }
-    participants.push(identity);
-    this.#phoneCanvassIdToParticipantDisplayName.set(
-      identity.activePhoneCanvassId,
-      participants,
-    );
+
+    const withId = PhoneCanvassCallerDTO.from({
+      ...propsOf(caller),
+      id: ++this.#nextId,
```
tdresser: Functional change here.

PR #260, opened 2025-11-01
Basic Twilio integration

tdresser:
There's a million tests I should probably write, but I think landing this first and iterating probably makes sense.

### DIFF
--- a/doc/schema/schema.json
+++ b/doc/schema/schema.json
@@ -822,7 +822,7 @@
   ],
   "driver": {
     "name": "postgres",
-    "database_version": "PostgreSQL 18.0 (Debian 18.0-1.pgdg13+3) on x86_64-pc-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit",
+    "database_version": "PostgreSQL 18.1 (Debian 18.1-1.pgdg13+2) on x86_64-pc-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit",
     "meta": {
       "current_schema": "public",
       "search_paths": [


grassroots-backend-e2etests/src/PhoneCanvass.e2e-spec.ts
```diff
@@ -12,6 +13,60 @@ import { UsersModule } from "grassroots-backend/users/Users.module";
...
+2daxkrhawl0vlez883m9s17m,4,4,,4,Kitchener Centre,400,Durham,Dally,Dietrich,en,2,1,,Dstreet,Rd,,,Leamington,N8H 2Q9,ON,5147382928,,,,5,,,phoning,,,,,,,,; ; vlu-43-02-ole-2022; fed-2021-sg; pre-2023-sg; plu-fed-2023; 2021-gpc-donor; donor-400+; import-2023-07-24; 2022-gpc-donor; true-multi-unit; import-2023-08-24; fed-2021-voted; 413-greenbelt; import-2023-10-24; active2-oct25; vlu-43c-01-ple-2023; volunteer-prospect; vote-pledge-2023; vlu-43c-03-pdl-2023; ac23-eday-vol-confirmed; ac23-wat-reg-vol; 2024-legalizeit-postcard; 2024-postelxn-textblast; pre-2024-sign-request; pre-2024-active3-vol; vlu-43-03-au-2024; ; ; 2023-advancepoll-voter; ac25voteadvance1; eday-vol-confirmed-2025; ; ; ; ; ; ; ;,d@d.com,`,
+  );
+  formData.append("voiceMailAudioFile", audioFile);
+  return formData;
+}
+
+async function updateTwilioCallStatus(
```
tdresser: As we build out more tests, I think we'll need a bunch of helpers like this to create a little DSL type thing.

grassroots-shared/src/dtos/PhoneCanvass/PhoneCanvass.dto.ts
```diff
@@ -141,14 +144,27 @@ export class PhoneCanvassAuthTokenResponseDTO extends createDTOBase(
...
+  // eslint-disable-next-line @darraghor/nestjs-typed/all-properties-are-whitelisted, @darraghor/nestjs-typed/all-properties-have-explicit-defined
+  @TwilioCallStatusDecorator()
+  CallStatus!: TwilioCallStatus;
+
   @IsOptional()
-  conference?: string;
+  CallDuration?: number;
+
+  /**
+   * Convert RFC 2822 timestamp to a ms since epoch
```
IanEdington: Is there a reason to convert to epoch? I find RFC 2822 much easier to read. Is it possible we would need timezone in the future?

grassroots-backend/src/phone-canvass/Scheduler/Strategies/NoOvercallingStrategy.ts
```diff
@@ -1,6 +1,8 @@
 import { filter, firstValueFrom } from "rxjs";
 import { PhoneCanvassSchedulerStrategy } from "./PhoneCanvassSchedulerStrategy.js";
+import { Injectable } from "@nestjs/common";
 
+@Injectable()
```
IanEdington: It doesn't look like this is needed

grassroots-backend/src/phone-canvass/PhoneCanvass.service.ts
```diff
@@ -26,18 +27,62 @@ import {
...
+} from "grassroots-shared/dtos/PhoneCanvass/CallStatus.dto";
+import { PhoneCanvassScheduler } from "./Scheduler/PhoneCanvassScheduler.js";
+import { Call } from "./Scheduler/PhoneCanvassCall.js";
+import { mergeMap } from "rxjs";
+import { PhoneCanvassSchedulerFactory } from "./Scheduler/PhoneCanvassSchedulerFactory.js";
 
 @Injectable()
 export class PhoneCanvassService {
   repo: EntityRepository<PhoneCanvassEntity>;
+  callsBySid = new Map<string, Call>();
```
IanEdington: I'm not seeing where calls are being removed from this map. I believe this would result in a memory leak. Same for `schedulers`.

The way we're doing things, it seems like calls need to stick around until the end of the phone canvass. Can we remove the phone canvass and associated calls when the phone canvass is done?

PR #261, opened 2025-11-04
Simulate callers joining, becoming ready, and becoming unready

tdresser:
This is the first step of simulating a call party.

We aren't simulating calls progressing, only callers joining and leaving.

### DIFF
--- a/grassroots-backend-e2etests/src/PhoneCanvass.e2e-spec.ts
+++ b/grassroots-backend-e2etests/src/PhoneCanvass.e2e-spec.ts
@@ -218,7 +218,6 @@ describe("PhoneCanvass (e2e)", () => {
           displayName: "Test",
           email: "Test@Test.com",
           activePhoneCanvassId: canvass.id,
-          ready: false,
         }),
       }),
     );


grassroots-frontend/src/Features/PhoneCanvass/Components/MarkReadyForCallsButton.tsx
```diff
@@ -30,7 +30,7 @@ export function MarkReadyForCallsButton(
   useEffect(() => {
     void (async (): Promise<void> => {
       await joinSyncGroup({
-        identity: new PhoneCanvassCallerDTO(),
+        identity,
```
tdresser: Not sure when I broke this...

Fixed now though!

grassroots-frontend/src/Features/PhoneCanvass/Logic/PhoneCanvassIdentityStore.ts
```diff
@@ -10,17 +10,19 @@ export interface PhoneCanvassCallerStore {
...
-          caller: undefined,
-          setCaller: (caller: PhoneCanvassCallerDTO): void => {
-            set({ caller });
-          },
-        };
-      },
-      {
-        name: "phonecanvass-caller-store",
-      },
+      persist(
```
tdresser: persist uses local storage to make the identity of a caller persist across page loads.

grassroots-backend/src/phone-canvass/PhoneCanvass.controller.ts
```diff
@@ -216,4 +220,16 @@ export class PhoneCanvassController {
     session.phoneCanvassCaller = caller;
     return caller;
   }
+
+  @Get("start-simulation/:id")
+  async startSimulation(@Param("id") id: string): Promise<VoidDTO> {
+    if (!(await getEnvVars()).ENABLE_PHONE_CANVASS_SIMULATION) {
+      throw new ForbiddenException(
+        "Can't simulate a phone canvass without ENABLE_PHONE_CANVASS_SIMULATION",
```
IanEdington: What is the benefit of this check? It seems like we would want to simulate in all environments.

IanEdington (approved):
This all looks pretty reasonable to me. I worry about having this much simulation code, and wonder if there's a way to more tightly match production, but I can't see how.

grassroots-backend/src/phone-canvass/PhoneCanvass.controller.ts
tdresser: Maybe? We really want to avoid hitting the simulate button on a real phone canvass, especially a partially completed one, as we could lose the state.

I think we could protect against this, but for now, it's kinda dangerous.

PR #273, opened 2025-12-03
Log env vars

tdresser:


grassroots-backend/src/GetEnvVars.ts
```diff
@@ -119,6 +126,35 @@ async function readSingleEnvironmentFile(
 // Cache environment variables.
 let envVariables: Environment | undefined = undefined;
 
+function logEnvVars(envVariables: Environment): void {
```
tdresser: Can I get a second pair of eyes on which should be sha1'd here?

rsalmond: LGTM! I don't see anything else that looks sensitive.

PR #280, opened 2025-12-11
Fix bugs

tdresser:


grassroots-backend/src/phone-canvass/Scheduler/PhoneCanvassMetricsTracker.ts
```diff
@@ -53,6 +57,9 @@ export class PhoneCanvassMetricsTracker {
       startWith(0),
       shareReplay({ bufferSize: 1, refCount: true }),
     );
+    // TODO: does forcing an early subscribe get this to happen earlier, avoiding an
```
tdresser: I have a better solution than this, but haven't had time to test it yet.

This does do the job!

PR #294, opened 2026-04-27
Windows compatibility

markwong167:
- Removed excess package.json
- added shared node_modules to dockerignore
- network_mode: host is incompatible with Windows, this fix _should_ be Linux compatible


markwong167:
It appears to be line 21 of .github/workflows/ci.yml where we need some PAT token.
> token: ${{ secrets.ACTION_PAT }}

Would it be better to use:
> token: ${{ github.token }}

tdresser (approved):
LGTM, I'd check with Ian for the token question.

